PR #25, opened 2015-09-09
empty files no longer allowed

jmasonherr:


boydgreenfield:
Note the ~35 byte limit is fairly arbitrary, but gives us a buffer on a reasonable size we'd ever expect to successfully process, since we need at least 1 k-mer plus a header separator, etc.

PR #36, opened 2016-12-20
Upload improvements

bovee:
This is a major overhaul of how we handle uploads and does the following:
 - Automatic (can be overridden with flags though) paired-end file interleaving
 - Streaming FASTX file validation and compression including current edge cases like tabs in header lines
 - Remove reliance on ~~`requests_toolbelt` and~~ `awscli` (replaced with `boto3`)


onecodex/cli.py
```diff
@@ -152,20 +156,70 @@ def samples(ctx, samples):
...
+        # "intelligently" find paired files and tuple them
+        paired_files = []
+        single_files = set(files)
+        for filename in files:
+            # convert "read 1" filenames into "read 2" and check that they exist; if they do
+            # upload the files as a pair, autointerleaving them
+            pair = re.sub('[._][Rr]1[._]', lambda x: x.group().replace('1', '2'), filename)
+            # we don't necessary need the R2 to have been passed in; we infer it anyways
+            if pair != filename and os.path.exists(pair):
+                if prompt:
```
boydgreenfield: Thoughts on moving this outside of the loop so we have a single prompt for _any_ paired files? I _think_ this is probably better from a UX perspective.

onecodex/cli.py
```diff
@@ -152,20 +156,70 @@ def samples(ctx, samples):
...
               default=False)
-@click.option("--max-threads", default=4,
-              help=OPTION_HELP['max_threads'], metavar="<int:threads>")
-@click.argument('files', nargs=-1, required=False,
-                type=click.Path(exists=True))
+@click.option('--max-threads', default=4,
+              help=OPTION_HELP['max_threads'], metavar='<int:threads>')
+@click.argument('files', nargs=-1, required=False, type=click.Path(exists=True))
+@click.option('--clean', is_flag=True, help=OPTION_HELP['clean'], default=False)
+@click.option('--interleave/--no-interleave', is_flag=True, help=OPTION_HELP['interleave'],
```
boydgreenfield: I prefer `--do-not-interleave` slightly. Thoughts? Also does using `--interleave` do anything? No right?

onecodex/cli.py
```diff
@@ -152,20 +156,70 @@ def samples(ctx, samples):
...
+                    single_files.remove(pair)
+                single_files.remove(filename)
+        files = paired_files + list(single_files)
+
+    if not clean:
+        warnings.filterwarnings('error', category=ValidationWarning)
+
+    try:
+        # do the uploading
+        if not no_threads:
```
boydgreenfield: Thoughts on just exposing a `--threads` param and having it default to max_threads, but be settable to 1? Slightly easier interface IMO.

onecodex/cli.py
```diff
@@ -152,20 +156,70 @@ def samples(ctx, samples):
...
+            ctx.obj['API'].Samples.upload(files, threads=max_threads)
+        else:
+            ctx.obj['API'].Samples.upload(files, threads=1)
+    except ValidationWarning as e:
+        sys.stderr.write('\nERROR: {}. {}'.format(
+            e, 'Running with the --clean flag will suppress this error.'
+        ))
+    except ValidationError as e:
```
boydgreenfield: Is this the gross warnings part? I think we can do `e.message` do shorten these up.

onecodex/cli.py
bovee: If we do that we can't explicitly say what files are being merged and we still have to go through the loop to determine if there are any files being interleaved...

onecodex/cli.py
bovee: I like this.

onecodex/cli.py
bovee: This is the much improved version. The remaining gross part is we have to do the `sys.stderr.write` to get the output formatted properly newline-wise (so we can't use `click`s notifications)

onecodex/cli.py
bovee: I just used the format from the `click` example; `--do-not-interleave` is a little long for my taste...

onecodex/cli.py
boydgreenfield: Ah. Cool!

onecodex/cli.py
boydgreenfield: Great let's update.

onecodex/cli.py
boydgreenfield: Let's get @sminot to weigh in here too, but I think better to not prompt every time. It's going to be awfully annoying to say "Yes" for each sample on a run. I worry this will just encourage careless use of `--force`/`--no-prompt`.

boydgreenfield:
Note that we're currently stripping out the header from the quality score line:

```
@SRR1649162.1 1 length=88
GTTCACCCCTAACGTTTCACCGACGATCAGCCAGCAGAATTGCTGCGAAAAGGGAAATGTGGGGCTAAACAGCGGCGACAGCGTCATC
+SRR1649162.1 1 length=88
<;51////*/<56/1/+/9>;>@B999999424;@@@B8;8<CCCCCHCC9;;89@:;<<CB@4?888/3773.,....44..33331
```

becomes:

```
@SRR1649162.1 1 length=88
GTTCACCCCTAACGTTTCACCGACGATCAGCCAGCAGAATTGCTGCGAAAAGGGAAATGTGGGGCTAAACAGCGGCGACAGCGTCATC
+
<;51////*/<56/1/+/9>;>@B999999424;@@@B8;8<CCCCCHCC9;;89@:;<<CB@4?888/3773.,....44..33331
```

Let's discuss if we want to do this or not. I'd lean towards not since: (1) it modifies the file; and (2) it should be very low overhead if compressed.

boydgreenfield:
OK I think this is about ready to merge. I'm going to test in a few different environments to ensure validation isn't incurring too much overhead, but after that LGTM @bovee!

bovee:
Okay, I just reviewed the whole thing and made a couple last-minute changes and I think this should be good to merge.

PR #57, opened 2017-08-14
Better viz

cindyzwang:
add heatmap and pca plotting


bovee:
Comments from Nick:

> A few high-level comments for thought / check in on tomorrow (all just suggestions):
> 1) Diversity functions should be verbs? `calc_simpson`? IDK about this. slightly odd as just `alpha_diversity` but  maybe not. What's the interface in skbio like? OK to mimic that I think.
> 2) "Jaccard distance" vs. "dissimilarity"? Does `skbio` call it `dissimilarity`. If so, gross but OK
> 3) `beta_diversity` and `alpha_diversity` should have default stats? (We can add this later if we're unsure _which_)
> 4) I'd like to move stuff out of `lib` and to top-level modules, so we can do `ocx.viz` or `ocx.helpers` vs. `ocx.lib.viz` which feels a bit nested/gross. Open to alternatives/pushback on this.
> 5) I suspect we'll want to review plotting stuff more and do a bit of work on default styles to give them an OCX "look and feel" but we can do this later (I haven't looked yet, just would be nice to tune up palettes etc.)
> 6) We probably have some test/coverage debt here.
> 7) I think we need to test some of this taxonomy code on messy/mixed/old databases.... may be a few edge cases in there we need to catch explicitly.
> 8) _This_ makes me wonder if we want `sentry` optionally enabled and on by default in our notebook environments so we can help people out... (should be off by default in the lib generally, but we could toggle with a custom env var)
> 9) _Later_ I think we need to come up with some rules for our import story (when it's in a function vs. top-level of the module) and probably add a `.tox` environment or two with the different deps.

bovee:
And, to touch on 7 above, the taxonomy tree generation seems to work against the oldest analysis I could find on Staging (against "Deprecated Staging DB") so I think this is fairly robust (at least the backend is coercing the results into something fairly robust...)

I added a (very simple) test anyhow to make sure we're at least somewhat covered on that.

boydgreenfield:
Closing in favor of #58.

PR #69, opened 2017-10-24
Fallow bugs

cindyzwang:
Upgrade Pandas to `0.20.3` and make a hack-y fix so it actually works how it's supposed to with `NaT` values.

### DIFF
--- a/onecodex/api.py
+++ b/onecodex/api.py
@@ -52,7 +52,7 @@ class Api(object):
         if api_key is None and bearer_token is None:
             try:
                 api_key = json.load(open(os.path.expanduser('~/.onecodex')))['api_key']
-            except:
+            except Exception:
                 pass
             if api_key is None:
                 api_key = os.environ.get('ONE_CODEX_API_KEY')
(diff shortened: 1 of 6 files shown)


onecodex/api.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, api_key=None,
         if api_key is None and bearer_token is None:
             try:
                 api_key = json.load(open(os.path.expanduser('~/.onecodex')))['api_key']
-            except:
+            except:  # noqa
```
boydgreenfield: We should make this `except Exception`

onecodex/utils.py
```diff
@@ -64,9 +64,9 @@ def valid_api_key(ctx, param, value):
     """
     Ensures an API has valid length (this is a click callback)"""
     if value is not None and len(value) != 32:
-        l = len(value)
+        length = len(value)
         raise BadParameter(
-            "API Key must be 32 characters long, not {}".format(str(l)))
+            "API Key must be 32 characters long, not {}".format(str(length)))
```
boydgreenfield: `.format(str(len(value)))`

boydgreenfield:
Lgtm!

PR #73, opened 2017-11-17
Adds support for filtering FASTQ files by Tax ID

### LINKED ISSUE #63: Add option to manually specify forward and reverse reads file
We currently auto-detect PE data, but do not have a mechanism for manually specifying forward and reverse files, e.g., `onecodex upload --forward <file1> --reverse <file2>`. We should add this.

### LINKED ISSUE #66: Failure on schema fetching if `simplejson` is installed
```
  File ".../site-packages/onecodex/api.py", line 169, in _fetch_schema
    client=self)
  File ".../site-packages/requests/models.py", line 884, in json
    self.content.decode(encoding), **kwargs
  File ".../site-packages/simplejson/__init__.py", line 533, in loads
    return cls(encoding=encoding, **kw).decode(s)
  File ".../site-packages/potion_client/converter.py", line 122, in __init__
    JSONDecoder.__init__(self, *args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'encoding'
```

Basically requests uses simplejson if installed which passes an `encoding` parameter through the Flask-Potion JSON decoding class to the `json` library (which doesn't support it). Possible fixes are:
 - submit a patch to use simplejson in Flask-Potion (if installed) 
 - monkey-patch the PotionJSONSchemaDecoder to drop the encoding argument if passed before initializing the `json` JSONDecoder

This may be an upstream flask-potion issue, but the way we're setting this up to handle our schema with all of our monkey-patching may not be a use-case they support? Some more investigation of why the other parts of `flask-potion` don't blow up when `simplejson` is installed is probably warranted.

### LINKED ISSUE #71: Add ability to filter a FASTQ by tax ID
The CLI should provide a flag that lets the user pass in the original FASTQ file for an `Analysis`, then filter reads from a FASTQ file that match a list of comma-delimited tax ID's.

clausmith:
This PR addresses the following:

- Adds support for specifying `--forward/--reverse` read files, closes #63 
- Fixes #66 by throwing an error message if `simplejson` is installed in the same environment
- Adds a `scripts` subcommand to allow for less-structured scripts to be included in the CLI
- Uses the above to add a script for filtering a given FASTQ file by a set of tax ID's, using a `Classification`; closes #71 

The FASTX filtering is implemented as part of a command subgroup that looks like:

`onecodex scripts filter_reads`

This will let us add more arbitrary workflows/scripts to the CLI going forward without cluttering the primary command namespace. The actual `filter_reads` command takes a `Classification` ID and the source FASTX file(s) and filters the read data based on a list of tax ID's.

```
$ onecodex scripts filter_reads --help
Usage: onecodex scripts filter_reads [OPTIONS] CLASSIFICATION FASTX

  Filters read files based on tax ID's

Options:
  -t, --tax-ids TEXT            A comma-delimited list of tax ID's to retain
                                [required]
  -r, --reverse PATH            The reverse (R2) read file, optionally
  --split-pairs / --keep-pairs  Keep only the read pair member that matches
                                the list of tax ID's
  -o, --out PATH                Where to put the filtered outputs
  -h, --help                    Show this message and exit.
```
 
The command supports a comma-separated list of tax ID's, as well as paired-end source files, and whether or not to split discordant pairs, when one read matches the chosen tax ID but not the mate.


clausmith:
This is just about wrapped, the last question is whether we really want to force quit if `simplejson` is installed, or just warn and continue.

onecodex/cli.py
```diff
@@ -60,6 +62,11 @@ def onecodex(ctx, api_key, no_pprint, verbose, telemetry):
     if verbose:
         log.setLevel(logging.INFO)
 
+    # Show a warning if simplejson is installed
```
boydgreenfield: Let's just open an issue to track fixing this upstream so we can remove this annoying warning (i.e., to make it work with `simplejson`).

onecodex/cli.py
```diff
@@ -80,6 +87,14 @@ def onecodex(ctx, api_key, no_pprint, verbose, telemetry):
         warn_if_insecure_platform()
 
 
+@onecodex.group('scripts', help='Assorted utility scripts')
+def scripts():
+    pass
+
+
+scripts.add_command(filter_reads.cli, 'filter_reads')
```
boydgreenfield: Great, this feels better than an `exec` call.

onecodex/cli.py
```diff
@@ -93,7 +108,7 @@ def analyses(ctx, analyses):
 @onecodex.command('classifications')
 @click.option("--read-level", 'readlevel', is_flag=True,
               help=OPTION_HELP['readlevel'])
-@click.option("--read-level-path", 'readlevel_path',
+@click.option("--read-level-path", 'readlevel_path', type=click.Path(),
```
boydgreenfield: Is this needed for the underlying `scripts`? Fine either way.

onecodex/cli.py
```diff
@@ -163,8 +188,16 @@ def samples(ctx, samples):
               default=True)
 @click.pass_context
 @telemetry
-def upload(ctx, files, max_threads, clean, no_interleave, prompt, validate):
+def upload(ctx, files, max_threads, clean, no_interleave, prompt, validate,
```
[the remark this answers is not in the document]
boydgreenfield: Let's still do this.

Second check becomes:
```
if forward and reverse:
    if len(files) > 2:
        click.echo('You may not upload more than 2 files when using the --forward and --reverse options.', err=True)
        sys.exit(1)
    files = [(forward, reverse)]
    no_interleave = True

File: onecodex/cli.py [hunk 5]
```diff
@@ -163,8 +189,16 @@ def samples(ctx, samples):
...
     """Upload a FASTA or FASTQ (optionally gzip'd) to One Codex"""
+
+    if (forward or reverse) and not (forward and reverse):
+        click.echo('You must specify both forward and reverse files', err=True)
+        sys.exit(1)
+    if forward:
+        files = [(forward, reverse)]
+        no_interleave = True
     if len(files) == 0:
         print(ctx.get_help())
```
boydgreenfield: `click.echo`?

onecodex/scripts/filter_reads.py
```diff
@@ -0,0 +1,174 @@
...
+    filtered_rows = []
+    for i, row in enumerate(results):
+        if row['Tax ID'] in tax_ids:
+            filtered_rows.append(i)
+        if i % 1000 == 0 and cb:
+            cb(1000)
+    return filtered_rows
+
+
+def get_record_count(fp):
```
boydgreenfield: Does calling `len` on the file object work? If so, would be simpler/cleaner.

onecodex/utils.py
```diff
@@ -169,18 +175,51 @@ def warn_if_insecure_platform():
...
+    """
+    Right now, potion-client is incompatible with requests when simplejson is
+    installed. Until this is patched upstream, we display this warning to
+    users when simplejson is installed.
+    """
+    m = ("\n"
+         "######################################################################################\n"  # noqa
+         "#                                                                                    #\n"  # noqa
+         "#  You currently have simplejson installed. Unfortunately, this library does not     #\n"  # noqa
+         "#  work properly alongside simplejson. Please install this library in a separate     #\n"  # noqa
```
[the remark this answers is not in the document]
boydgreenfield: Let's still add "or uninstall simplejson"

onecodex/cli.py
clausmith: It's not required, I just noticed and thought it might make sense to add the type.

onecodex/scripts/filter_reads.py
clausmith: No, we'd have to convert the file into a list first. But that involves reading the whole file into memory, which I think is sub-optimal since some of these FASTQ's can be GB's. This is just a memory-efficient line counter.

onecodex/cli.py
boydgreenfield: Nope that's fine!

onecodex/scripts/filter_reads.py
boydgreenfield: That's fine. I think we have a `len` method on our other file which actually reads it through for use with `requests-toolbelt`. Note we should double check that `seek(0)` works properly for our iterators (though I'm sure your tests cover this).

clausmith: Looks like the `len` on the `FASTXTranslator` uses the length of the `Buffer` in bytes, whereas here's we're checking line count. In any case we can definitely move/refactor this in the next round.

PR #91, opened 2018-05-01
Bc wip tag upload

### LINKED ISSUE #87: Allow metadata and tags to be set at upload time
We should allow then when uploading _only one file_ (or two files that get interleaved). 

Idea for format: `onecodex upload sample.fastq --tag isolate --tag healthy --metadata key=value` (or `key:value`), where all keys in the metadata _default_ fields flow through to those default fields, and other keys create custom metadata.

bcamarda:
closes #87 

This pull enables a user to set tags and metadata from the CLI as part of the upload process.

TODO - update the Readme.


.pytest_cache/v/cache/nodeids
```diff
@@ -0,0 +1,142 @@
+[
```
boydgreenfield: Let's add to `.gitignore`

.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
boydgreenfield: Let's add to `.gitignore`

onecodex/cli.py
```diff
@@ -186,12 +186,26 @@ def samples(ctx, samples):
...
 
+    tag_array = []
+    if tags:
+        tag_array = [tag.strip() for tag in tags.split(',')]
+
+    metadata_opts = {}
+    if metadata:
+        for metadata_kv in metadata:
+            split_metadata = metadata_kv.split('=')
+            if len(split_metadata) > 1:
```
boydgreenfield: We should probably warn the user via `stderr` if they improperly invoke the CLI like `onecodex upload $SAMPLE --metadata field` with no value.

onecodex/cli.py
```diff
@@ -186,12 +186,26 @@ def samples(ctx, samples):
...
+                metadata_value = '='.join(split_metadata[1:])
+                metadata_opts[split_metadata[0].lower()] = metadata_value
```
boydgreenfield: We should consider forcing metadata field names into snakecase (we do this within the One Codex web app).

onecodex/cli.py
```diff
@@ -264,6 +280,80 @@ def upload(ctx, files, max_threads, clean, no_interleave, prompt, validate,
         sys.exit(1)
 
 
+def update_sample_tags_and_metadata(sample_uuids, ctx, tag_array, metadata_opts):
```
boydgreenfield: Probably makes sense to move these functions to some other internal module / helpers file.

onecodex/models/__init__.py
```diff
@@ -79,7 +79,10 @@ def __getattr__(self, key):
                 else:
                     if key == 'id':
                         # undo the bad coercion from potion_client/resource.py#L111
-                        return str(value)
+                        if value is None:
```
boydgreenfield: 👍

onecodex/lib/upload.py
```diff
@@ -115,6 +115,7 @@ def progress_bar_display(file_id, bytes_transferred, validation=False):
     progress_bar = None if log_to is None else progress_bar_display
 
     # first, upload all the smaller files in parallel (if multiple threads are requested)
+    uploading_uuids = []
```
boydgreenfield: 👍

tests/test_cli.py
```diff
@@ -170,10 +170,12 @@ def test_logout_creds_dne(runner, mocked_creds_path):
     (["temp.fa", "temp2.fa"], True),
 ])
 def test_standard_uploads(runner, upload_mocks, files, threads):
```
boydgreenfield: Would be nice to have a few test cases `@pytest.mark.parametrize` around uploading various metadata keys/values and coercion of those into the right types/key names. This could be as tests of the internal functions I'm suggesting we move into a `helpers.py` file somewhere above.

PR #95, opened 2018-05-15
Bc metadata and tag upload

### LINKED ISSUE #84: Variable log messages while uploading
When upload a list of files, sometimes we provide the uploaded sample ID and sometimes we don't. We should always output the sample ID once the upload has finished.

### LINKED ISSUE #87: Allow metadata and tags to be set at upload time

### LINKED ISSUE #90: Unsaved models' IDs are improperly coerced into String "None"
This is causing issues when attempting to create Tags from the CLI, and is probably causing additional problems.

bcamarda:
Adds Metadata and Tag support to the CLI.
Closes #84 
Closes #87 
Closes #90 
Closes #91


onecodex/lib/upload.py
```diff
@@ -59,7 +59,7 @@ def _wrap_files(filename, logger=None, validate=True):
 
 
 def upload(files, session, samples_resource, server_url, threads=DEFAULT_UPLOAD_THREADS,
-           validate=True, log_to=None):
+           validate=True, log_to=None, metadata=None, tags=[]):
```
boydgreenfield: Change to `None`

PR #113, opened 2018-09-06
Handle lack of write access to the home directory

bovee:
This can cause breakage in some restricted environments (like heavily permission docker images).

### DIFF
--- a/onecodex/api.py
+++ b/onecodex/api.py
@@ -191,7 +191,11 @@ class ExtendedPotionClient(PotionClient):
 
             # always resave the creds (to make sure we're removing schema if we need to be or
             # saving if we need to do that instead)
-            json.dump(creds, open(creds_fp, mode='w'))
+            try:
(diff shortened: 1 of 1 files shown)


polyatail:
Do we want this to say something aloud? Could be better especially now that login is required for most functions.

boydgreenfield:
@polyatail I think that's a good idea, and would be in favor of logging a message to STDERR that says something along the lines of `We were unable to save your credentials to ~/.onecodex. You'll need to pass your API key directly using `--api-key` to use the CLI.`

Or something like that. Then happy to merge this.

polyatail:
There's also the case where ~/.onecodex exists and isn't readable, which breaks a lot of stuff...

boydgreenfield:
@polyatail Is this ready for final review and merging?

polyatail:
Superseded by PR #123

Issue #118, opened 2018-11-02
Snake case metadata names by default when uploading


boydgreenfield:
In our frontend uploader, we snake case metadata field names. Via the CLI or `ocx.Samples.upload` method, however, we do not. This leads to inconsistent naming, e.g.,:

![image](https://user-images.githubusercontent.com/535969/47890701-4c3ff080-de0d-11e8-8d88-d7d054868286.png)

We should consider automatically snake casing these names in our library directly (and perhaps enforcing these on the API side).

polyatail:
I think it probably does this in the CLI here:

```python
            if len(split_metadata) > 1:
                metadata_value = split_metadata[1]
                appendables['metadata'][snake_case(split_metadata[0])] = metadata_value

    appendables = validate_appendables(appendables, ctx.obj['API'])
```

Where on the API handles this?

Issue #117, opened 2018-10-18
Add better error message when trying to upload anonymously


boydgreenfield:
This is not a good error message. We should catch and prompt for login directly. May be worth having a `login_required` or similar decorator that we can put on individual click functions to do this check and prompt as needed.

```
10:34:12 (ocx) {master} ~/OC/create-react-app-antd$ onecodex upload ~/Downloads/MSA-1000.16S.short.fastq 
No One Codex API credentials found - running anonymously
Uploading: Preparing upload(s)...    
ERROR: 'HTTPError' object does not support indexing
Please feel free to contact us for help at help@onecodex.com
```

polyatail:
I addressed this by adding a login_required decorator in cli.py and requiring logins for all functions but login/logout.

boydgreenfield:
@polyatail Let's discuss, we don't _technically_ require that a user is logged in to use the API, e.g., they can access public samples by ID, but in practice that's probably not that useful via the CLI.

polyatail:
Hmm ok... I added back in the anonymous login for non-upload functions. But stuff like 'analyses' and 'samples' comes back empty.

boydgreenfield:
@polyatail As discussed, I like your requirement on everything but `login` and `logout`.

Issue #114, opened 2018-09-25
Add `--include-children` option to filter reads script


boydgreenfield:
This would be very useful for, e.g., fetching all reads for a given species _and all of its strains_.

boydgreenfield:
Note we should also add a `--include-parents` flag to get things _at or above_ a given tax ID too.

@clausmith preference on `--with-children`/`--with-parents` vs. `--include-children`/`--include-parents`? Former is shorter so perhaps better.

polyatail:
Question: for --with-parent, is the desired behavior to produce only the direct ascendants (if that's a thing) of the specified tax_id? For E. coli, for example, Bacteria, Proteobacteria, Gammaproteobacteria, Enterobacteriales, Enterobacteriaceae, Escherichia, E. coli and no others.

boydgreenfield:
@polyatail Yes I think so. The idea here is that those are all of the reads that could presumably map back to a given genome via alignment or some orthogonal method.

Issue #83, opened 2018-02-23
Add support for an exclusive mode in filter_reads


clausmith:
Right now, the `filter_reads` script only _includes_ reads with the given tax IDs. We should add an option to instead _exclude_ reads matching these tax IDs.

Also, we should add a flag to create a filtered FASTQ without host reads.

clausmith:
@polyatail this is implemented in the `additional-filtering` branch. Just needs tests and rebasing.

PR #123, opened 2018-11-09
Require authentication and add features to filter_reads()

polyatail:
This pull request closes several issues and makes a few related changes:

- #117 Prints a better error message when trying to upload anonymously by requiring authentication for every function but login/logout
- #118 Simple fix of a bug that caused metadata names to be non-uniformly cased, by moving snake_case() call out of cli.upload and into sample.upload 
- #114 Add functionality to scripts.filter_reads, allowing all child descendants of the specified `--tax-id` to be included in the filter
- #83 Allows scripts.filter_reads functionality to be reversed, excluding all matches (rather than including them)
- Added some new JSON classifications to the unit test suite, to allow testing of `--with-children` that requires a matching JSON and TSV
- #124 Relocated some code from cli.py to models.Samples.upload() to search for projects by name or project_name (with a little error handling). Allows users to specify projects via the API, previously only accessible via CLI.
- #113 Explicit handling of IO errors related to accessing credentials file (~/.onecodex). Merged _login and _silent_login private functions and issue either warnings.warn() from API or click.echo() from CLI if the file doesn't exist, can't be parsed by `json` or has a permissions problem.

See this new usage:

```
(venv) mewtwo:onecodex as$ onecodex scripts filter_reads
Usage: onecodex scripts filter_reads [OPTIONS] CLASSIFICATION_ID FASTX

  Filter a FASTX file based on the taxonomic results from a
  CLASSIFICATION_ID

Options:
  -t, --tax-id TEXT             Filter reads mapping to tax IDs. May be passed
                                multiple times.  [required]
  --with-children               Keep reads of child taxa, too. For example,
                                all strains of E. coli
  -r, --reverse PATH            The reverse (R2) read file, optionally
  --split-pairs / --keep-pairs  Keep only the read pair member that matches
                                the list of tax ID's
  --exclude-reads               Rather than keep reads matching reads,
                                choosing this option will exclude them.
  -o, --out PATH                Where to put the filtered outputs
  -h, --help                    Show this message and exit.
Try "onecodex scripts filter_reads -h" for help.

Error: Missing argument "CLASSIFICATION_ID".
```

This PR is ready to be merged after review.


.circleci/config.yml
```diff
@@ -95,8 +95,8 @@ jobs:
           name: Run flake8
           command: |
             . venv/bin/activate
-            flake8 --ignore E501 --exclude onecodex/schemas/* onecodex/
-            flake8 --ignore E501 --exclude onecodex/schemas/* tests/
+            flake8 --ignore E501,W504,W605 --exclude onecodex/schemas/* onecodex/
```
boydgreenfield: Let's fix these linting issues rather than adding an ignore. Fix for `W605` is to use raw strings, e.g., `rb'\d'`. Fix for `W504` is to probably rewrite some of those lines with `.format()` which is better style anyhow.

`W605` is important in particular as it looks like this behavior will break Python 3.8 once that's released.

onecodex/cli.py
```diff
@@ -92,14 +103,36 @@ def scripts():
     pass
 
 
-scripts.add_command(filter_reads.cli, 'filter_reads')
+@scripts.command('filter_reads', help='Filter a FASTX file based on the taxonomic results from a CLASSIFICATION_ID')
```
boydgreenfield: Do we have to move this in order for `login_required` to work? Ideally, we'd leave these and import them as the old code does. Otherwise, this file will become _very_ verbose as we add more and more scripts over time.

onecodex/scripts/filter_reads.py
```diff
@@ -53,20 +53,61 @@ def write_fastx_record(record, handler):
...
-@click.option('-r', '--reverse', type=click.Path(), help='The reverse (R2) '
-              'read file, optionally')
-@click.option('--split-pairs/--keep-pairs', default=False, help='Keep only '
-              'the read pair member that matches the list of tax ID\'s')
-@click.option('-o', '--out', default='.', type=click.Path(), help='Where '
-              'to put the filtered outputs')
-@click.pass_context
-@pretty_errors
-def cli(ctx, classification_id, fastx, reverse, tax_id, split_pairs, out):
+def make_taxonomy_dict(classification, parent=False):
```
boydgreenfield: This is good/fine for now, _but_ we should give some thought to a `onecodex.taxonomy` module and refactoring this to use that. 

It'd be nice to have, e.g., a `Taxonomy` class that could be instantiated from a variety of sources (our results here, an NCBI dmp file set, perhaps our internal "build.json" networkx serialization format, etc.) and provide methods/properties like:
- `children`
- `parents`
- `lowest_common_ancestor(a, b, ...)`

etc. 

See github.com/onecodex/taxonomy for a not-very-good and old implementation fo this idea that I think we should integrate into this library directly. Only additional dependency should be `networkx` I'd guess.

onecodex/scripts/filter_reads.py
```diff
@@ -53,20 +53,61 @@ def write_fastx_record(record, handler):
...
+    Takes a classification data frame returned by the API and parses it
+    into a dictionary mapping a tax_id to its children (or parent).
+    Restricted to tax_id's that are represented in the classification
+    results.
+    """
+
+    tax_id_map = {}
+
+    if parent:
+        for row in classification.table().itertuples():
```
boydgreenfield: Subtle: It would be better to use `classification.results()` and iterate through those as the `.table()` method imports pandas which we do _not_ install by default. 

100% of CLI commands should work with a vanilla `pip install onecodex` vs. a full library install `pip install onecodex[all]`.

As needed we should also test this command in a tox environment _without_ all dependencies so we can track these. That may be worth setting up as part of this PR.

onecodex/scripts/filter_reads.py
```diff
@@ -53,20 +53,61 @@ def write_fastx_record(record, handler):
...
+            if row.parent_tax_id is not None and \
+               row.tax_id is not None:
+                tax_id_map[row.tax_id] = row.parent_tax_id
+    else:
+        for row in classification.table().itertuples():
```
boydgreenfield: Same as above.

tests/test_scripts.py
```diff
@@ -5,14 +5,29 @@
...
 
 from onecodex import Cli
+from tests.test_cli import make_creds_file
 
 
-@pytest.mark.parametrize('paired,split_pairs', [
-    (False, False),
-    (True, True),
-    (True, False),
+@pytest.mark.parametrize('paired,split_pairs,with_children,exclude_reads', [
```
boydgreenfield: Awesome. Such parametrization! ;)

onecodex/cli.py
polyatail: The reason I did this originally is because I put the login_required decorator in cli.py, which imports scripts/filter_reads.py. In order to then access the login_required decorator from scripts/filter_reads.py, I'd have to circularly import login_required from cli.py.

polyatail: I was able to revert back to the scripts.add_command() style by moving login_required into auth.py and importing auth.login_required() in both filter_reads.py and cli.py.

onecodex/scripts/filter_reads.py
polyatail: Fixed.

onecodex/scripts/filter_reads.py
polyatail: Fixed.

onecodex/cli.py
polyatail: Though now, an Api() instance is returned in both cli.py and in auth.py, which doesn't seem very clean. Perhaps cli.login (which doesnt really do anything except call auth._login) should be refactored.

.circleci/config.yml
polyatail: Okey dokey, I re-enabled the warnings and fixed them. W605 was mostly adding an additional escape character--adding 'r' or 'rb' seemed to break things. W504 was mostly changing the way long strings were split across multiple lines.

onecodex/cli.py
polyatail: If you want to see what I mean, checkout cli.login and note that it creates an Api() instance only to compare its e-mail address to the e-mail address returned by auth._login. I'm thinking that code can be moved into auth somewhere.

onecodex/lib/upload.py
```diff
@@ -31,7 +31,7 @@ def interleaved_filename(filename):
     if re.match('.*[._][Rr][12][_.].*', filename[0]):
         return re.sub('[._][Rr][12]', '', filename[0])
     else:
-        print('Does not match')
+        print('WARNING: Paired-end filenames do not match--are you sure they are correct?')
```
boydgreenfield: This should probably be `click.echo`?

boydgreenfield:
PR LGTM! Only questions are:
1. There’s a print that should probably be `click.echo`? Or maybe not? (If it's in library code vs. CLI)
2. Does `click` interop ok with the `warnings.warn`? just not sure if we should use that vs. something `click`-y.

@polyatail You should feel free to merge the PR and address the above two issues later _or_ tweak and then merge. Your call. Thanks for the PR!

onecodex/lib/upload.py
polyatail: True...

polyatail:
I used warnings.warn() in api.py just because click isn't imported there (since Api() is not necessarily called from the command line). But ultimately I think they are both just writing to stderr unless we change their parameters.

Issue #79, opened 2018-01-19
`filter_reads` command should be aware of filtered reads


boydgreenfield:
Currently, the `filter_reads` script doesn't pay attention to if reads are filtered or not. This is probably generally fine, as most reads filtered out are for specific organisms end-users won't select, but we should _probably_ default to only using filtered reads if filtering information is available and then make a `--include-filtered-reads` or `--include-removed-reads` flag for including both.

polyatail:
Is this the same as --exclude-reads in #83 ?

boydgreenfield:
@polyatail No, this is more subtle – and terrible written (I had to read my comment 2x to understand). We have a filtered tax ID column in our TSVs, and we "filter" out likely erroneous results and report them as 0. We should consider defaulting to that column when available.

bcamarda:
Maybe we should rename this command `subset_reads` to make things simpler...

PR #141, opened 2018-12-04
Major changes to filter_reads

polyatail:
This pull request addresses #79 by:

* Ignoring taxonomic assignments not passing filter
* Letting user override this functionality by passing --include-lowconf

In addition, this PR might break existing code by:

* Producing output that's different because of ignoring assignments not passing filter
* Renaming filter_reads to subset_reads


tests/test_scripts.py
```diff
@@ -1,30 +1,43 @@
...
 
 
-@pytest.mark.parametrize('paired,split_pairs,with_children,exclude_reads', [
+@pytest.mark.parametrize('paired,split_pairs,with_children,exclude_reads,include_lowconf', [
     # single end
-    (False, False, False, False),
-    (False, False, True, False),  # --with-children
-    (False, False, True, False),  # --exclude-reads
-    (False, False, True, True),   # --with-children --exclude-reads
+    (False, False, False, False, False),
```
boydgreenfield: This is insane, but OK 😱

onecodex/scripts/subset_reads.py
```diff
@@ -0,0 +1,387 @@
...
+@click.option('-r', '--reverse', type=click.Path(),
+              help='The reverse (R2) read file, optionally.')
+@click.option('--validate/--no-validate', is_flag=True, default=True,
+              help='By default, validate FASTQ records (slow). Disable to drastically increase speed')
+@click.option('--with-children', default=False, is_flag=True,
+              help='Match child taxa of those given with -t (e.g., all strains of E. coli)')
+@click.option('--split-pairs', default=False, is_flag=True,
+              help='By default, if either read in a pair matches, both will match. Choose this '
+                   'option to consider each paired-end read separately. Resulting files may *not* '
+                   'have the same number of reads!')
```
boydgreenfield: Let's call it `--subset-pairs-independently` and have the following help message:

```
By default, if either read in a pair matches, both will be retained in the subset file. With this option, R1 and R2 files will be evaluated independently. Note that the subset output FASTQs are *not* guaranteed to have the same number of reads!
```

polyatail: Changed as requested.

tests/test_scripts.py
polyatail: It's just well-tested.

Issue #140, opened 2018-12-02
Error trying to upload sample with unicode in filename


clausmith:
I get the following error trying to upload a file with a `ö` in the filename:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/ubuntu/.local/lib/python2.7/site-packages/onecodex/lib/upload.py", line 152, in _wrapped   
    wrapped_args[-1].value = '{}'.format(e)
  File "<string>", line 11, in setvalue
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 28: ordinal not in range(128)
```

boydgreenfield:
Yea... let's fix this. I'd lean towards providing a useful error message, but at the moment I think it's probably best for us to enforce ASCII-only filenames (umlauts will inevitably break many downstream tools too).

A y/n prompt to convert their filename to ascii could also be OK if there's a reasonable way to do that (I'm not sure offhand, but I'd guess most of these errors will be `[éèüå]` -> `[eeua]` kind of issues.

bovee:
To be clear, this looks like a `bytes`/`str` issue in the client that we should probably fix and then separately warn/error in the upload proxy/server-side on unicode characters too (although I'm not thrilled about disallowing every non-English file name).

boydgreenfield:
@bovee I think disallowing non-ASCII filenames for now is a reasonable and probably necessary requirement. The problem is that lots of tools will break on these filenames, and so it just opens a huge set of broken pipeline edge cases. 

The _best_ solution is probably to address this via sample deduplication -- i.e., rename the file to a sanitized internal filename (could even be a UUID), expose the original filename for downloads/exports and subsequent analyses. Let's discuss more.

PR #157, opened 2018-12-21
Warn user and rename files with accents

polyatail:
* Adds `unidecode` as dependency, can convert chinese characters to ascii, too!
* Added tests and filenames with accents
* Small tweaks for python2 compat


bovee:
I'm not seeing the link set on the title, so for reference this is for issue #140 .

polyatail:
In that case, should we have some flag for renaming to ASCII? Or make the users do it?

Issue #161, opened 2019-01-03
FASTQ file with -- lines silently gets truncated


boydgreenfield:
A FASTQ file with `--` on lines will silently get truncated by the validator rather than failing as it should. Example file:

```
@HEADER
GCGGAATCGCCTTCCTTCCCGGGCAACAGTCGCATCGTT
+
JJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJ
--
@HEADER2
GCGGAATCGCCTTCCTTCCCGGGCAACAGTCGCATCGTT
+
JJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJ
``` 

Will only upload a 1 read file.

boydgreenfield:
Note this is caught by the proxy, so we can decide whether to fix in the CLI or just handle via the proxy:

![image](https://user-images.githubusercontent.com/535969/50648821-0f899600-0f31-11e9-9d08-370b09a13d87.png)

polyatail:
Hmm... my two cents, we should fix this in the validator and pin a release ASAP, then perhaps shoot for refactoring the uploader and switching over to the proxy 100% of the time for CLI uploads in the next 1-2 sprints.

polyatail:
Here are some potential ways of solving this issue in the CLI:

- Switch from `match()` to `finditer()` which will find all FASTQ records in the buffer, skipping bad lines
- Count number of lines in the file. For FASTQ, this should be divisible by 4 and the result of that division should equal the number of records we parse.
- Parse file as groups of four lines, checking for a FASTQ regexp (or other) match for each.

PR #162, opened 2019-01-03
Do a better job of catching invalid file uploads

polyatail:
This addresses #161 by checking if there's leftover data in the buffer that can't be parsed by our FASTX regexps. There's also the case where half a record is leftover in the buffer, which will be completed once more data is read. If the record *still* isn't completed after more data is read from the file and appended to the buffer, raise.

This branch also pins a release at v0.3.1.


onecodex/lib/inline_validator.py
```diff
@@ -249,7 +251,15 @@ def __iter__(self):
             while True:
                 match = self.seq_reader.match(self.unchecked_buffer, end)
                 if match is None:
+                    if (waiting_for_data or eof) and len(self.unchecked_buffer) - end > 0:
+                        raise ValidationError('Sequence in buffer does not match expected format.')
```
boydgreenfield: Let's have this say "Your FASTA/Q file terminates abruptly or is otherwise malformed"

Issue #160, opened 2019-01-03
__setattr__ should be aware of ResourceList types


polyatail:
Presently, if the property of a model is set to a `ResourceList`, the model can not be saved. For example:

```
>>> samples[0].tags = samples[1].tags
>>> samples[0].save()
TypeError: Object of type 'ResourceList' is not JSON serializable
```

This is because `__setattr__` is not aware of the `ResourceList` type and does not properly pull the underlying resource out and assign it to the model.

Issue #159, opened 2019-01-03
Return lists of Classifications, Analyses, and Samples as SampleCollection


polyatail:
Presently, multiple instances of these models are returned from a `where` query in a list. They should instead be returned as a `SampleCollection` that behaves like a list. A `SampleCollection` will make available additional methods for analyzing classification results.

- [X] ~~In a minimal install~~ (existing code requires pandas), users can call `to_otu()` to obtain a BIOM-formatted table, or perhaps an option of the `results` method?
- [X] In a full install, users can call methods from these modules: `taxonomy`, `viz`, and `distance` which are are added as mixins.
- [X] The object must be written such that it functions like a list, perhaps by subclassing `ResourceList`
- [X] It should not fail if importing e.g., `pandas` is unsuccessful. In this case, it should just not add the additional methods.
- [X] The `results` method of a SampleCollection should return a subclassed `pd.DataFrame` that carries metadata and taxonomy dfs along for the ride.
- [X] `ResultsDataFrame` should allow access to the same analysis methods as a `SampleCollection` via the `ocx` namespace.

Issue #158, opened 2019-01-02
Slices of ResourceList objects should be ResourceList objects


polyatail:
Presently, `ResourceList` slices are returned as single instances of the resource type if the slice includes only a single item. For example:

```
>>> type(sample[1].tags)
ResourceList
>>> type(sample[1].tags[0])
onecodex.models.misc.Tags
```

However, when a slice is taken that includes multiple items, a `list` is returned. It should be another `ResourceList`:

```
>>> type(sample[1].tags)
ResourceList
>>> type(sample[1].tags[:2])
list
```

Issue #152, opened 2018-12-15
Adopt new docstring conventions and update docstrings


polyatail:
We should choose from NumPy or Google style docstrings and update/improve existing ones.

polyatail:
Add comments to everything and enable linting.

PR #164, opened 2019-01-08
Visualization overhaul (2 of 2)

polyatail:
The PR is a substantial overhaul to how `Samples` and their associated `Classifications` are analyzed in a notebook environment.

The following issues were addressed:
- #74 Scatter plots of distances between samples (calculated using a distance metric, e.g., unifrac) can now be visualized with principal component analysis, via two algorithms: deterministic pcoa, or smacof, an iterative optimization. This is exposed to the user via `plot_mds`.
- #130 Enabled easier lookup of Tags resources only, rather than a change affecting all models.
- #145 To facilitate rendering of report PDFs (via HTML and weasyprint), setting an environment variable will enable the onecodex renderer in altair, which produces high-res PNGs.
- #149 In `plot_heatmap`, the samples and taxa are ordered according to euclidean distance.
- #150 When clusters are generated, in `plot_heatmap` or in `plot_distance`, the user can now select the clustering algorithm, e.g., single- or average-linkage.
- #152 New methods and most older methods have updated Numpy-style docstrings now.
- #158 Fixed bug where ResourceList slices were returned as lists, when they should have been ResourceLists.
- #159 Lists of `Samples` or `Classifications` are now returned wrapped in a new `SampleCollection` object which exposes lots of additional analyses methods that operate on the objects contained within.
- #160 Fixed bug where properties of a model which were `ResourceList` objects were not being saved.

New functionality has been added as part of the `SampleCollection` object, some examples of which are outlined below.

**Retrieving a project and generating plots**
All plotting functions are now available as methods of the new `SampleCollection` object.

```python
>>> project1 = ocx.Projects.get('d53ad03b010542e3')
>>> samples1 = ocx.Samples.where(project=project1.id, public=True, limit=10)
>>> type(samples1)
onecodex.models.SampleCollection
>>> samples1.plot_pca(color='Bacteroides')
```

**Obtaining a DataFrame OTU table**
We introduce a new subclassed `ResultsDataFrame` which is returned when the user requests an OTU table from a `SampleCollection`. This object transfers taxonomy and metadata along with it, which is automatically subset to only ever contain the taxonomy and metadata of taxa present in the `ResultsDataFrame` and metadata belonging to `Samples` present.

```python
results = samples1.results(rank='phylum', normalize=False)

# take the fourth root and plot a heatmap
(results ** 0.25).ocx.plot_heatmap()

# restrict to taxa with 100 reads or more and run PCA
results.loc[:, results.sum() > 100].ocx.plot_pca(color='Bacteroidetes')

# do PCoA with weighted Unifrac
results.ocx.plot_mds(method='pcoa', metric='weighted_unifrac', color='Bacteroidetes')
```

Metadata and taxonomy can be accessed two ways:

```python
samples1.metadata == samples1.results().ocx.metadata
samples1.taxonomy == samples1.results().ocx.taxonomy
```

The values in the `ResultsDataFrame` can be manipulated as usual. Analysis methods which are available to `SampleCollection` are also available in the `ResultsDataFrame.ocx` namespace via `pd.api.extensions`.

**Manipulating contents of SampleCollection**
The `Samples` in a `SampleCollection` can be manipulated as if they were a list. A user might notice that one sample is "bad" and want to exclude it from their analyses. Or they might want to combine multiple projects together, or analyze their own project in the context of a larger public dataset.

```python
project2 = ocx.Projects.get('be33784611fe4c19')
samples2 = ocx.Samples.where(project=project2.id, public=True, limit=10)

# analyze these two projects together
combined = samples1 + samples2

# exclude this sample
del combined[4]

combined.plot_pca(size='Firmicutes')

# get a combined OTU table and write to a CSV file
combined.results(rank='genus', normalize=True).to_csv('otu_table.csv')

# and in long format
combined.results(table_format='long').to_csv('otu_table.long.csv')
```


onecodex/taxonomy.py
```diff
@@ -1,77 +1,98 @@
...
+        for tax_id in tax_ids:
+            tax_ids_to_keep.append(tax_id)
+            tax_ids_to_keep.extend([x.name for x in tree.find(tax_id).ancestors()])
+
+        tree = tree.copy()
+        tree.remove_deleted(lambda n: n.name not in tax_ids_to_keep)
+
+        return tree
+
+    def tree_prune_rank(self, tree, rank='species'):
```
bovee: Can we add a comment with an example here, like "with `rank='species'` all strain nodes will be removed, but all genus, etc nodes will be kept" (I assume that's what this is doing).

onecodex/viz/_distance.py
```diff
@@ -1,118 +1,326 @@
...
+            r_squared = []
+
+            for axis in [0, 1]:
+                mds_dist = pos.copy()
+                mds_dist[::, axis] = 0
+                mds_dist = squareform(euclidean_distances(mds_dist).round(6))
+                r_squared.append(pearsonr(mds_dist, squareform(dists))[0])
+
+            # label the axes
+            x_extra_label = 'r² = %.02f' % (r_squared[0],)
```
bovee: I'm fine with the unicode, but could these be LaTeX instead? I think matplotlib can handle that, but not sure about other backends. This may prevent problems down the road with source file encodings.

requirements.txt
```diff
@@ -2,31 +2,31 @@
 boto3>=1.4.2
 click>=6.6
 jsonschema>=2.4
+python-dateutil>=2.5.3
```
bovee: I'm fairly reluctant to pin with a `>=`; we should probably pin to specific minor-release versions of packages, e.g. using ~=, to prevent unexpected downstream breakages from taking out this library.

tox.setup.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py34
+envlist = py27,py36
```
bovee: Why aren't we supporting py34 and py35 any more? How hard would it be to support them?

onecodex/helpers.py
```diff
@@ -1,169 +1,328 @@
...
+class ResultsDataFrame(pd.DataFrame):
+    _metadata = ['ocx_rank', 'ocx_field', 'ocx_taxonomy', 'ocx_metadata']
+
+    def __init__(self, data=None, index=None, columns=None, dtype=None, copy=False, ocx_data={}):
+        self.ocx_rank = ocx_data.get('ocx_rank', None)
+        self.ocx_field = ocx_data.get('ocx_field', None)
+        self.ocx_taxonomy = ocx_data.get('ocx_taxonomy', None)
+        self.ocx_metadata = ocx_data.get('ocx_metadata', None)
+
+        pd.DataFrame.__init__(self, data=data, index=index, columns=columns, dtype=dtype, copy=copy)
```
audy: I'm not sure what this line is doing. `__init__` has to return `None` and this DataFrame isn't being assigned to anything.

audy:
Here are some comments/suggestions.

## Separate normalization steps from stats.

I think that keeping all preprocessing/normalization steps separate from statistics steps would be clearer and also allow the user to perform the preprocessing once and then pass the pre-processed analysis object to multiple stats functions

```python
# perform normalization

processed_analysis = normalize(agglomerate(analysis_object, rank='Phylum'))

# do some stats
alpha = alpha_diversity(processed_analysis)

beta = beta_diversity(processed_analysis)
```

## Since the entire API is changing I think it's worthwhile to evaluate a few possible API designs

Proposal A: method chaining. This would be pretty familiar to anyone used to phyloseq (or d3). I also think it reads well and is easy to type:

```python
experiment.agglomerate('Phylum').normalize().beta_diversity()
```
Proposal B: IDK what this is called but it's familiar to scikit-learn which is an API for processing DataFrames and doing stats/machine learning:

```python
transformer = ReadCountNormalizer(method="proportion")

summary_statistic = ShannonDiversityIndex()

stats = summary_statistic(transformer.transform(results_data_frame))
```
This API design can be made a little bit more readable using something like [`Pipeline`](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html)

```python
# straight up ripoff of scikit-learn
my_analysis = Pipeline([
  ('transform', ReadCountNormalizer),
  ('top5', SelectTopTaxa(n=5)),
  ('alpha_diversity', ShannonDiversityIndex())
])

my_analysis.transform(results_data_frame)
```

It would even be possible to have multiple APIs. Again these are just suggestions but I think it's worthwhile. Also, it's possible to implement these APIs on top of what you've built already (and have multiple ones at the same time).

## "ResultsDataFrame" is kind of ambiguous? I'm bad at naming things though. I am thinking that there are going to be multiple things named "results" in users' notebooks and it might get confusing.

## Even though the odds of a false positive are probably nil, a good alternative would be to store a bool on `ResultsDataFrame` and have any normalization function turn it to `True`.

https://github.com/onecodex/onecodex/blob/4e7a7692d80ed43d4b212c991f9f34ec8e3c8ef6/onecodex/helpers.py#L107-L111

onecodex/helpers.py
```diff
@@ -1,169 +1,328 @@
...
+            Return only the top N most abundant taxa.
+        threshold : `float`, optional
+            Return only taxa more abundant than this threshold in one or more samples.
+        remove_zeros : `bool`, optional
+            Do not return taxa that have zero abundance in every sample.
+        normalize : {'auto', True, False}
+            Convert read counts to relative abundances (each sample sums to 1.0).
+        table_format : {'long', 'wide'}
+            If wide, rows are classifications, cols are taxa, elements are counts. If long, rows are
+            observations with three cols each: classification_id, tax_id, and count.
```
audy: Would it be possible to split up all of these transformation steps into separate public functions? It would be easier to test and I think it's clearer (although the style you've chosen _is_ more similar to Pandas' conventions).

onecodex/helpers.py
```diff
@@ -1,169 +1,328 @@
...
+        # subset by taxa
+        if rank:
+            tax_ids_to_keep = []
+
+            for tax_id in df.keys():
+                if self.taxonomy['rank'][tax_id] == rank:
+                    tax_ids_to_keep.append(tax_id)
+
+            if len(tax_ids_to_keep) == 0:
+                raise OneCodexException('No taxa kept--is rank ({}) correct?'.format(rank))
```
audy: Are the rank names stored anywhere? It might be better to assert that the rank is valid early.

audy:
(playing around with the lib now. I'll give you some thoughts on my first impressions)

https://github.com/onecodex/onecodex/blob/4e7a7692d80ed43d4b212c991f9f34ec8e3c8ef6/onecodex/models/__init__.py#L238

I think `@property` doesn't work well here because it's really a function that calls an API (and changes the objects state). As a user, I expect that this will return immediately when it's really doing 900 GET requests.

I think it would be better to drop the internal caching mechanisms and just use something external and explicit like https://github.com/reclosedev/requests-cache which automatically caches all HTTP requests and stores them in an sqlite3 database. As a user, I am unaware of when things are being pulled from the API or from the cache. This way, I am in control of the cache. We could make it simpler for the user (like `onecodex.enable_cache()`).

onecodex/helpers.py
polyatail: It's equivalent to `super(ResultsDataFrame, self).__init__()` in Python2-style syntax, or `super().__init__()` in Python3 style. The reason I used this instead of `super()` is that ultimately `ResultsDataFrame` will inherit from multiple classes and I wanted to be explicit about which parent class's `__init__` I wanted to call.

onecodex/helpers.py
```diff
@@ -1,169 +1,328 @@
...
 
-    classifications = [c for c in classifications if c.success is True]
 
-    # we'll fill a dict that eventually turn into a pandas df
-    df = {
-        'classification_id': [c.id for c in classifications]
-    }
+class AnalysisMethods(VizPCAMixin, VizHeatmapMixin, VizMetadataMixin, VizDistanceMixin):
+    def __init__(self, results, metadata, taxonomy, field):
+        self._cached = {'results': results,
```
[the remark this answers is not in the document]
polyatail: Yes, these methods work on `ResultsDataFrame` and also `SampleCollection`. Both of those objects inherit from the `AnalysisMethods` class.

onecodex/distance.py
```diff
@@ -1,120 +1,119 @@
...
-
-def cityblock(classifications, field='readcount_w_children', rank='species'):
-    """Compute the Manhattan distance between two classifications."""
-    assert field in ACCEPTABLE_FIELDS
-    counts, tax_ids, ids = beta_counts(classifications, field=field, rank=rank)
-    return skbio.diversity.beta_diversity('cityblock', counts, ids)
+from onecodex.taxonomy import TaxonomyMixin
+
+
+class DistanceMixin(TaxonomyMixin):
```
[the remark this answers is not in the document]
polyatail: Hmm, the reason for the refactor here is in anticipation of more taxonomy functions which will probably make it into a different type of subclassed `pd.DataFrame`, or perhaps `skbio.tree.TreeNode`

onecodex/viz/_distance.py
polyatail: I wish =\ Altair's manual told me to just use unicode. Should we change it to `r-squared` instead?

tox.setup.ini
polyatail: If we do, I'd like to add them as a separate test and also test 3.6. There was some issue (don't recall exactly) that worked in 3.4 but not in 3.6.

polyatail:
IIRC `primary_classifications` doesn't actually trigger any API calls, but I get what you're saying.

audy:
`SampleCollection.primary_classification` checks `self._cached` and if the result is False, dispatches `_classification_fetch` which will trigger an API request at least in the case of the model being `Classifications`.

audy:
Updated `SampleCollections.primary_classifications` to only cache the results after all classifications have been fetched to prevent a potential gotcha.

tox.setup.ini
boydgreenfield: I'd like to test on 2.7, 3.4, and 3.6 and/or 3.7. There are definitely people who will be pinned on 3.4 for a long time. <3.4 was kind of a mess if I recall correctly, and I think we're OK only supporting 2.7 and 3.4+.

Issue #155, opened 2018-12-18
Add bash completion support


boydgreenfield:
This would be very fancy and nice. Probably not necessary to autocomplete IDs (too much), but autocompleting all of our, e.g., `scripts` commands would be slick.

polyatail:
Seems like this might already exist be default with `click`--see [here](https://click.palletsprojects.com/en/5.x/bashcomplete/). Will look more into how to activate it.

bcamarda:
Look into what other packages do to enable bash completion (and figure out documentation?)

polyatail:
Bash completion requires each command to have its own script which tells it how to do tab completion. Some systems will have a `bash-completion` package installed, which provides such scripts for many common tools. These scripts are located in either `/usr/local/etc/bash_completion.d` or `/etc/bash_completion.d`. The default `.bash_profile` or `.bashrc` scripts on a system may or may not source completion scripts from this directory.

```
# mac
brew install bash-completion

# linux
apt-get install -y bash-completion
```

The polite way to enable completion for `onecodex` would be to insert our completion script into this directory, if it exists and is writeable. Otherwise, it's on the user to install that package and our completion script, or install it directly into their `.bash_profile`.

This can be accomplished by adding a short bit of code that's run post-install in `setup.py`. We might also add something to the documentation in case someone wants to add it to their own `.bash_profile`.

Issue #154, opened 2018-12-18
Automatically detect `ONE_CODEX_BEARER_TOKEN` in CLI usage too


boydgreenfield:
We currently autodetect this when instantiating the client, but not via the CLI. This would allow more seamlessly using the CLI in a notebook environment.

Issue #153, opened 2018-12-17
Send exception message to Sentry any time we sys.exit(1)


boydgreenfield:
Currently, in code like this:

```
    except (ValidationError, UploadException, Exception) as e:
        # TODO: Some day improve specific other exception error messages, e.g., gzip CRC IOError
        sys.stderr.write('\nERROR: {}'.format(e))
        sys.stderr.write('\nPlease feel free to contact us for help at help@onecodex.com\n\n')
        sys.exit(1)
```

We lose valuable info to Sentry, making debugging errors harder. We should do a `sentry.captureException` call in spots like this and _perhaps_ revert our special `SystemExit` handling.

boydgreenfield:
So we do send the message... but maybe the exception is more/equally useful? TBD.

bcamarda:
See: https://sentry.onecodex.com/onecodex/one-codex-cli/issues/3009/?query=is:unresolved

polyatail:
A couple things:

- Most times we're calling `sys.exit()`, we can and should probably be calling `ctx.exit()` since we're within a Click context.
- Other times we're calling `sys.exit()`, we completely handle the error and I think it's appropriate that we're making that call. They are mostly for authentication failures or permissions errors accessing `~/.onecodex`. I don't think Sentry needs to know about these. Given that, I think it's appropriate to revert the  `SystemExit` catch in the `@telemetry` decorator.
- There is only one time that we call `sys.exit` after catching an exception that we don't entirely handle: the one @boydgreenfield quoted above for `UploadError` or `Exception` while uploading a file. In this case, I think we should just let the exception happen, get reported to Sentry, and let the user see it. Any `ValidationError` raised in this case is pretty well handled, so those I think we should continue to call `sys.exit` after and not send to Sentry.

boydgreenfield:
@polyatail I think it's still useful from a telemetry POV to get the `ValidationError` info, because we can't guarantee that our validators are always working properly. That said, this is kind of a moot point when we move to a proxy. We could send it as a message though, or categorize those types of issues within the Sentry web interface. Thoughts?

PR #167, opened 2019-01-09
Minor auth and sentry changes, bash completion

polyatail:

- #153 Removes capturing `SystemExit` calls and is more choosey about when to `sys.exit()`
- #154 Allows CLI to authenticate via `API_KEY` or `BEARER_TOKEN` environment variables
- #155 Adds a post-install script which tries to enable bash completion if the user has a `bash-completion` package installed in a writeable way

Low priority, can go into a larger release.

### DIFF
--- a/onecodex/auth.py
+++ b/onecodex/auth.py
@@ -171,22 +171,40 @@ def _logout(creds_file=None):
 
 
 def login_required(fn):
-    """
-    Decorator for the CLI for requiring login before proceeding.
+    """Requires login before proceeding, but does not prompt the user to login. Decorator should
+    be used only on Click CLI commands.
+
+    Notes
+    -----
+    Different means of authentication will be attempted in this order:
+        1. An API key present in the Click context object from a previous successful authentication.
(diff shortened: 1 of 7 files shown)


onecodex/auth.py
```diff
@@ -171,22 +171,40 @@ def _logout(creds_file=None):
...
+        2. A bearer token (ONE_CODEX_BEARER_TOKEN) in the environment.
+        3. An API key (ONE_CODEX_API_KEY) in the environment.
+        4. An API key in the credentials file (~/.onecodex).
     """
 
     @wraps(fn)
     def login_wrapper(ctx, *args, **kwargs):
-        if 'API_KEY' in ctx.obj and ctx.obj['API_KEY'] is not None:
-            ctx.obj['API'] = Api(api_key=ctx.obj['API_KEY'], telemetry=ctx.obj['TELEMETRY'])
+        api_kwargs = {'telemetry': ctx.obj['TELEMETRY']}
```
audy: Looks good but I think we should document the precedence that is used to set the api key b/c I've had issues with packages before where I could not figure out how the API key was determined.

Issue #151, opened 2018-12-15
Allow users to group samples in heatmap by metadata


polyatail:
This would appear as two horizontally concatenated heatmaps, one for each group of sample, clustered in one go such that rows in both charts can be labeled with one set of labels.

Issue #136, opened 2018-12-01
Build dendogram primitive


bcamarda:

polyatail:
This should also include a function to cluster a ResultsDataFrame by one or both axes.

PR #172, opened 2019-01-14
Allow grouping of heatmap elements by metadata

polyatail:

#136 I broke the box plot and dendrogram plotting functions out into their own `viz/_primitives.py` file. This will make it easier to re-use these bits of code.
#151 By user request, `plot_heatmap` now takes a categorical metadata field (or tuple containing multiple metadata fields), and performs clustering (with user-specified metric and linkage) on a per-category basis.

This code will generate clustered heatmaps at each taxonomic rank:

```python
ranks = ['species', 'genus', 'family', 'class', 'order', 'phylum']
plts = {}

for r in ranks:
    plts[r[0]] = samples.plot_heatmap(
        metric='weighted_unifrac', linkage='median',
        haxis='fish', threshold=0.1, rank=r,
        return_chart=True
    )

plts['p'] & plts['o'] & plts['c'] & plts['f'] & plts['g'] & plts['s']
```

In the below example plot, the value of metadata field 'fish' is reported above the heatmap.

image


boydgreenfield:
Why does this decrease test coverage in `_heatmap.py` by 50%? Can we add a test to address that?

image

boydgreenfield:
@polyatail Works for me!

Issue #109, opened 2018-08-16
Scope changes in upload threading


bovee:
This code is a little confusing and hard to debug (there was some confusion on issue #106). If this isn't giving us a significant perf boost, it's possible we want to remove it.

- [ ] Decide whether to remove upload threading from CLI

polyatail:
Previously, we were uncompressing, validating/interleaving, and recompressing files locally. This meant  it was more likely to be CPU-bound during the upload process. Now that we are blindly interleaving (or in the case of single-ended files, sending them as-is), we're more likely to be IO-bound or to hit the upstream bandwidth cap of the user. The code for threading isn't that substantial, but it may not be necessary. I could go either way.

Issue #108, opened 2018-08-16
Remove multipart upload code


bovee:
We should handle everything through our upload proxy so we can remove this code path.

Issue #105, opened 2018-08-16
Remove file upload validation code


bovee:
We now have a streaming proxy that can do validation across all of our endpoints so we probably don't need to do this in the CLI/Python API client.

polyatail:
To do this, I had to change `filter_reads` to use `skbio.io`. This may not be a long-term solution as skbio has already dropped support for python 2 as of v5.0.0.

PR #179, opened 2019-01-17
Refactor uploads via CLI

polyatail:
This PR addresses #105, #108, and #109 by:

- Removing FASTX validation code and deleting `inline_validator.py`
- No longer distinguishing between small and large uploads--all go through same code path
- Continue to interleave FASTX files when paired via `FASTXInterleave` object
- `FASTXInterleave` decompresses when necessary, interleaves, and uploads uncompressed
- Upload single-ended files via `FASTXPassthru` which leaves files as-is
- Switch `filter_reads.py` to `skbio.io` FASTX parser, having removed our own parsers

In addition, these changes were made:

- Progress bar during upload now uses `click.progressbar`
- Errors in multithreaded upload are printed as they happen rather than at the end
- Passing `-v` sets loglevel to DEBUG rather than INFO
- `pytest` upgraded to 3.3.0
- Tests using old `make_creds_file()` were switched over to `mocked_creds_file` fixture


boydgreenfield:
I think this is our checklist of TODOs before removing the validation code:
- [ ] Add a `retry_url` or similar param in cases where our proxy fails with a 500 (which _will_ not have quite as good uptime as S3, hard as we try)
- [ ] Add a hook to call `/samples/<id>/cancel_upload` when we cancel a pending upload
- [ ] Fix any remaining proxy edge cases that may be blocking (see our internal proxy repo)

Issue #171, opened 2019-01-12
Improve testing of plots produced by viz


polyatail:
Presently, tests succeed if the plotters don't raise. Obviously, it would be better if we check that the plots produced contain the correct information. Since we're using Vega, this is fairly simple as we can check the JSON returned by Altair.

Also, tests in `viz`, `taxonomy`, and `distance` should all be given a once-over to make sure redundant tests are minimized. Presently, they are thorough but slow.

Issue #168, opened 2019-01-11
Improve manipulation and filtering of Tags


polyatail:
The use of `Tags` objects to filter `Samples` would be powerful, but isn't well supported right now. It would be nice if some of this worked:

```python
>>> Tags(name='isolate') in sample.tags
True
>>> 'isolate' in sample.tags
True
>>> my_samples = Samples.where(tag=Tags(name='isolate'))
>>> my_samples = Samples.where(tag='isolate')
```

We should decide if one or both of the above syntaxes should be implemented. `Tags` don't really have anything other than a name, so perhaps its reasonable to support strings. To get "in" working, we need to add `__hash__` to `onecodex.models.Tags`. Querying by tags will require modifying the bespoke where method, `onecodex.models.Samples.where`.

PR #191, opened 2019-01-24
Improve test coverage and more intuitive Tags

polyatail:

- #168 Can now use `onecodex.Samples.where(tags='isolate')` style syntax to return a `SampleCollection` of samples matching a certain tag.
- #171 Improve test coverage and depth for viz, upload, and others

### DIFF
--- a/onecodex/models/sample.py
+++ b/onecodex/models/sample.py
@@ -6,7 +6,7 @@ import warnings
 
 from onecodex.exceptions import OneCodexException
 from onecodex.lib.upload import upload
-from onecodex.models import OneCodexBase, Projects
+from onecodex.models import OneCodexBase, Projects, Tags
 from onecodex.models.helpers import truncate_string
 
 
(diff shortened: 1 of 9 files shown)


onecodex/models/sample.py
```diff
@@ -24,6 +24,11 @@ def where(cls, *filters, **keyword_filters):
         public = keyword_filters.get('public', False)
         instances_route = 'instances' if not public else 'instances_public'
         limit = keyword_filters.get('limit', None if not public else 1000)
+        tag = keyword_filters.pop('tag', None)
```
boydgreenfield: So, we actually support this in the API and shouldn't need to get all samples and filter in memory. Here's an example public URL:

```
https://app.onecodex.com/api/v1/samples?where={"tags": {"$contains": {"$ref": "/api/v1/tags/dbcf5b98bca54a16"}}}
```

Consequently, I think the best way to implement this is to fetch the tag by name if we have a non-UUID (and maybe have some kind of `is_uuid()` function to skip this check) and then include it in the `.where` query directly.

Note we also implement a custom `$containsany` and `$containsall` selector in our backend that allows passing an array of tags and doing more complex intersects. We should look at the performance of that query code before leaning on it too heavily, but those may be useful too.

onecodex/models/sample.py
```diff
@@ -75,6 +80,23 @@ def where(cls, *filters, **keyword_filters):
...
+                    if tag in s.tags:
+                        samples_to_keep.append(s)
+                except HTTPError:
+                    # this is here for some strange error where a sample has a tag that doesn't exist?
+                    pass
+
+                if len(samples_to_keep) == limit:
+                    break
+
+            samples = samples_to_keep
```
boydgreenfield: See comment above re: removing and simplifying this logic (plus shipping less data!).

tests/test_api_models.py
```diff
@@ -81,6 +89,13 @@ def test_resourcelist(ocx, api_data):
         tags1.append(sample)
     assert 'object of type' in str(e.value)
 
+    if sys.version_info.major < 3:
+        with pytest.raises(AttributeError):
```
boydgreenfield: Just out of curiosity, why does this raise? Probably worth a comment for future us.

polyatail: Apparently lists in Python 2 do not have a clear() method

onecodex/models/sample.py
polyatail: OK, this was a lot simpler. Changed as requested.

onecodex/models/sample.py
```diff
@@ -25,6 +25,40 @@ def where(cls, *filters, **keyword_filters):
...
+            tags = [tags]
+
+        new_tags = []
+
+        for t in tags:
+            if isinstance(t, Tags):
+                new_tags.append(t)
+            else:
+                # is it a uuid?
+                if len(t) == 16:
```
boydgreenfield: Let's write an `is_ocx_id` helper function that does this plus and alphabet check. Something like:

```
assert set(x for x in name).issubset(string.hexdigits)
```

PR #208, opened 2019-02-06
Add bargraph to Viz

audy:
- [x] plots a bargraph
- [ ] ~merge non top-n taxa into a single category (so that bar graphs add up to 1)~
- [x] consistency with the rest of the `viz` module
- [x] testing
- [x] documentation
- [ ] ~add support for `haxis` argument?~ moved to #215
- [ ] ~smart bar sorting: (if plotting at genus rank, sort bars by family first etc...)~
- [ ] ~xmax should not exceed 1.0 (or 100% of total reads)~

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -1,8 +1,8 @@
 from onecodex.exceptions import OneCodexException
-from onecodex.viz import VizPCAMixin, VizHeatmapMixin, VizMetadataMixin, VizDistanceMixin
+from onecodex.viz import VizPCAMixin, VizHeatmapMixin, VizMetadataMixin, VizDistanceMixin, VizBargraphMixin
 
 
-class AnalysisMixin(VizPCAMixin, VizHeatmapMixin, VizMetadataMixin, VizDistanceMixin):
+class AnalysisMixin(VizPCAMixin, VizHeatmapMixin, VizMetadataMixin, VizDistanceMixin, VizBargraphMixin):
     """Contains methods for analyzing Classifications results.
 
     Notes
--- a/onecodex/viz/_bargraph.py
+++ b/onecodex/viz/_bargraph.py
@@ -1,0 +1,130 @@
+import altair as alt
+from onecodex.exceptions import OneCodexException
+
+
+class VizBargraphMixin(object):
+    """Plot a bargraph of relative abundance of taxa for multiple samples
+
+    Parameters
+    ----------
+    rank : {'auto', 'kingdom', 'phylum', 'class', 'order', 'family', 'genus', 'species'}, optional
+        Analysis will be restricted to abundances of taxa at the specified level.
+    normalize : 'auto' or `bool`, optional
+        Convert read counts to relative abundances such that each sample sums to 1.0. Setting
+        'auto' will choose automatically based on the data.
+    return_chart : `bool`, optional
+        When True, return an `altair.Chart` object instead of displaying the resulting plot in
+        the current notebook.
+    top_n : `int`, optional
+        Display the top N most abundant taxa in the entire cohort of samples.
+    threshold : `float`
+        Display only taxa that are more abundant that this threshold in one or more samples.
+    title : `string`, optional
+        Text label at the top of the plot.
+    xlabel : `string`, optional
+        Text label along the horizontal axis.
+    ylabel : `string`, optional
+        Text label along the vertical axis.
+    tooltip : `string` or `list`, optional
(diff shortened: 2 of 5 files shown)


onecodex/analyses.py
```diff
@@ -177,6 +177,9 @@ def to_df(self, rank='auto', top_n=None, threshold=None, remove_zeros=True, norm
         table_format : {'long', 'wide'}
             If wide, rows are classifications, cols are taxa, elements are counts. If long, rows are
             observations with three cols each: classification_id, tax_id, and count.
+        include_not_top_n : { True, False }
```
polyatail: Did we end up not  using this kwarg?

onecodex/viz/_bargraph.py
```diff
@@ -0,0 +1,124 @@
+from onecodex.exceptions import OneCodexException
+import altair as alt
```
polyatail: Let's try to keep the imports in two groups: xeno-modules at the top, then those from within `onecodex`, with an empty line between the two groups. And put them in alphabetical order. I might have messed this up elsewhere myself.

onecodex/viz/_bargraph.py
```diff
@@ -0,0 +1,124 @@
...
+            Text label at the top of the plot.
+        xlabel : `string`, optional
+            Text label along the horizontal axis.
+        ylabel : `string`, optional
+            Text label along the vertical axis.
+        tooltip : `string` or `list`, optional
+            A string or list containing strings representing metadata fields. When a point in the
+            plot is hovered over, the value of the metadata associated with that sample will be
+            displayed in a modal.
+    """
```
polyatail: Can the docstring contents be inline with """ ?

onecodex/viz/_bargraph.py
```diff
@@ -0,0 +1,124 @@
...
+
+        for f in tooltip:
+            df[magic_fields[f]] = magic_metadata[magic_fields[f]][
+                df["classification_id"]
+            ].tolist()
+
+        # add sample filenames
+        df["display_name"] = self.metadata["_display_name"][
+            df["classification_id"]
+        ].tolist()
```
polyatail: `_metadata_fetch` has a secret keyword where the string "Label" will be converted to the contents of `_display_name`. The benefit here is that the x axis title will then be "Label" if `xlabel=None`. Maybe add "Label" to `tooltip` and get rid of this?

onecodex/viz/_bargraph.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # add taxa names
+
```
polyatail: y u put newline here tho

onecodex/viz/_bargraph.py
```diff
@@ -0,0 +1,124 @@
...
+        # about taxonomic ranks but as all taxonomic data will be coming from
+        # OCX this should be okay)
+        #
+
+        chart = (
+            alt.Chart(df)
+            .mark_bar()
+            .encode(
+                x=alt.X("display_name", axis=alt.Axis(title=xlabel)),
+                y=alt.Y("readcount_w_children", axis=alt.Axis(title=ylabel)),
```
polyatail: Let's check if `xlabel` or `ylabel` are `None` and provide a default behavior? I don't remember... if we set `title=None` does that mean there's no axis labels?

onecodex/viz/_heatmap.py
```diff
@@ -41,6 +41,8 @@ def plot_heatmap(self, rank='auto', normalize='auto', top_n='auto', threshold='a
             A string or list containing strings representing metadata fields. When a point in the
             plot is hovered over, the value of the metadata associated with that sample will be
             displayed in a modal.
+        legend: `string`, optional
+            Variable to map the color scale to (default is 'abundance')
```
polyatail: 👍

onecodex/viz/_bargraph.py
audy: Labels will be the same as if `axis=` wasn't specified. Altair is inconsistent with this. For example, setting `chart.properties(title=None)` will result in an error.

screen shot 2019-02-15 at 11 23 26 am

onecodex/analyses.py
audy: I was going to add this feature but decide that I should not include it in this PR. Thanks for catching.

onecodex/viz/_bargraph.py
audy: set default `xaxis="Sample"`

PR #249, opened 2019-04-15
Speed up import times

boydgreenfield:
We could also explore using [`apipkg`](https://pypi.org/project/apipkg/)


polyatail:
Is this ready to go for the next release?

Issue #257, opened 2019-05-13
Improve documentation and handling of new tag creation


boydgreenfield:
Currently, creating a new tag is rather awkward and requires the following code:

```
t = ocx.Tags(name="my new tag")
t.sample = my_sample
t.save()
```

We should improve this to allow passing a `sample` param into `ocx.Tags()` directly or via another mechanism (e.g., allowing a tag to be appended into a `sample.tags` list and then handling that save). We should also include more explicit direction on how to create new tags in the docs string.

Issue #253, opened 2019-05-08
Expose sorting functionality in plot helpers


clausmith:
While it's possible to re-encode the `Altair` object with sorting information after creation, it would be nice to have a `sort_{x,y}` and `order_{x,y}` kwargs that can be passed to the `plot_*` helpers to control sorting for things like bar charts.

PR #263, opened 2019-06-13
Allow sorting of axes in plotters, improve Tags creation API

polyatail:
Addresses #253, #257.


polyatail:
I think goals for the next release should be to 1) improve performance of subset_reads (or even just make it default to the faster parser), 2) make start-up time faster by moving the imports around.

PR #278, opened 2019-07-19
Allow user to turn off saving (uploading) SVGs/PNGs in notebooks

polyatail:
This PR addresses an issue with large notebooks in the notebook service. Rendered images have to be uploaded back to the notebook service to save a `.ipynb` and this can hang indefinitely and/or result in partially saved/exported notebooks.

```python
from onecodex.viz import renderer_settings

# disable saving images to notebooks
renderer_settings(png_or_svg=False)
```

Users can switch between PNGs (smaller), SVGs, or saving no images using this new settings function. If no images are saved, they will be rendered on the server using node and the `vega-cli` package.

### DIFF
--- a/onecodex/notebooks/exporters.py
+++ b/onecodex/notebooks/exporters.py
@@ -7,6 +7,7 @@ import json
 from nbconvert.exporters.html import HTMLExporter
 import os
 import pytz
+import subprocess
 from traitlets import default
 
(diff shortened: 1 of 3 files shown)


boydgreenfield:
For clarity for those looks at this down the line, the syntax is actually `renderer_settings(png=False, svg=False)`.

PR #286, opened 2019-07-23
Add support for experimental models

polyatail:
This PR enables access to `AnnotationSet`, `Assembly`, `Genome` and `Taxon` models when the appropriate kwarg is passed to `Api()`

### DIFF
--- a/onecodex/api.py
+++ b/onecodex/api.py
@@ -36,6 +36,7 @@ class Api(object):
         base_url=None,
         telemetry=None,
         schema_path="/api/v1/schema",
+        **kwargs
     ):
 
         if base_url is None:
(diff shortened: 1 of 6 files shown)


onecodex/models/darwin.py
```diff
@@ -0,0 +1,101 @@
+from onecodex.models import OneCodexBase, ResourceList
```
polyatail: Rename to `experimental.py`

onecodex/api.py
```diff
@@ -43,6 +43,9 @@ def __init__(
             if base_url != "https://app.onecodex.com":
                 warnings.warn("Using base API URL: {}".format(base_url))
 
+        if schema_path != "/api/v1/schema":
```
polyatail: Add `**kwargs` and check if `experimental` in kwargs, warn about not being supported, and switch schema_path to experimental

PR #294, opened 2019-07-26
Use authentication while fetching API schema

polyatail:
Allows restriction of experimental API features on the backend.


boydgreenfield:
@polyatail This doesn't cause a regression w/r/t weird not getting "bad API key" messages because of a 401/403 on the schema fetch, does it? If not, this LGTM. But let's just double check that.

polyatail:
Works ok for me in testing, as long as `cache_schema=False` is set when `experimental=True`

boydgreenfield:
Cool! Feel free to merge and pin then!

Issue #298, opened 2019-08-07
`atexit` race condition bug on KeyboardInterrupt


boydgreenfield:
We had a user report the following stack trace, triggering an unexpected message from the `atexit` handler. Notably this isn't in Sentry so it looks like we may also not be capturing events in Raven during `atexit` handling (we need more exhaustive testing on these if so).

```
Finalizing...  [####################################]  100%             

example1.fastq: finished as sample XYZ
example2.fastq: finished as sample ABC

^C

Aborted!

Canceling...  [####################################]  100%              

Canceled upload for sample: XYZ

Error in atexit._run_exitfuncs:

Traceback (most recent call last):
  File "/usr/local/lib/python3.4/site-packages/onecodex/lib/upload.py", line 556, in cancel_atexit
    samples_resource.cancel_upload({"sample_id": fields["sample_id"]})
  File "/usr/local/lib/python3.4/site-packages/onecodex/vendored/potion_client/links.py", line 123, in __call__
    response, response_data = self.make_request(data, params)
  File "/usr/local/lib/python3.4/site-packages/onecodex/vendored/potion_client/links.py", line 99, in make_request
    self.raise_for_status(response)
  File "/usr/local/lib/python3.4/site-packages/onecodex/vendored/potion_client/links.py", line 90, in raise_for_status
    raise HTTPError(http_error_msg, response=response)
requests.exceptions.HTTPError: {'message': 'Upload already successful', 'success': False}
```

PR #299, opened 2019-08-09
Docstring linting, path autocompletion, upload cancel exception catching

polyatail:
This PR:

- #152 Adds automatic linting of docstrings in NumPy format and updates existing docstrings to pass.
- Note that I'm currently ignoring D100-105 which would force us to add docstrings all over the place, and D202 which requires no blank line after docstrings in function definitions. D202 conflicts with `black` linting.
- #298 Catches an exception when trying to cancel an upload that's already succeeded upon Ctrl+C
- Bumps version of Click to 7.0.0 which should fix a bug calculating ETA when `progressbar.pos = 0`
- Enables tab autocompletion of filenames and directories. It's not the greatest since we're kind of limited by Click's API but it's better than nothing IMO.
- Idk how this came to be, but we were re-using the flask-potion session with our API for uploads via proxy. Sets up a separate session for each upload, which should fix any connection pool errors as a side effect.
- Limits concurrent uploads to 8.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ lint:
 	flake8 onecodex/
 	flake8 tests/
 	black --check -l 100 --exclude vendored/* onecodex/ tests/
+	pydocstyle --convention=numpy --add-ignore=D100,D101,D102,D103,D104,D105,D202 --match-dir='[^vendored].*' onecodex/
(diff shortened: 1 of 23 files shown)


boydgreenfield:
@polyatail I'll review in the AM, but awesome!

PR #301, opened 2019-08-14
Support for sample preuploads

Keats:
Only makes sense to be merged after the backend API changes are merged and deployed.


boydgreenfield:
These changes are now live on our backend, _but_ we should also ensure that if the upload ends up being a multipart upload we ensure that works too (this is related to upload retries). I'll discuss with @polyatail and we'll get this merged on Monday.

PR #308, opened 2019-08-29
Add support for organization=True ocx.Samples.where filter. Closes #307.

### LINKED ISSUE #307: Add support for searching organization samples in `ocx.Samples.where`
Probably with the flag `organization=True` as with `public=True`.

boydgreenfield:


onecodex/models/sample.py
```diff
@@ -19,11 +19,43 @@ def __repr__(self):
...
+
+        Parameters
+        ----------
+        limit : `int`, optional
+            If set, retrieve a maximum of `limit` samples. Note a limit of 1000 samples is automatically
+            enforced for queries of all public samples (`public=True`)
+        organization : `bool`, optional
+            If True, search all samples within your organization (including your samples)
+        public : `bool`, optional
+            If True, search all public samples (limited to 1000 results)
```
Keats: it should mention `organization` and `public` are exclusive of each other, you can't look for samples either public *or* within the org

onecodex/models/sample.py
```diff
@@ -19,11 +19,43 @@ def __repr__(self):
...
+        organization = keyword_filters.pop("organization", False)
+        instances_route = "instances"
+        if organization is True:
+            instances_route = "instances_organization"
+        if public is True:
+            instances_route = "instances_public"
+
+        # Set default filters
+        keyword_filters["_instances"] = instances_route
+        keyword_filters.setdefault("limit", None if not public else 1000)
```
Keats: not using the negative would be a bit clearer: `1000 if public else None`. Or set `limit = None` at the top and set it to 1000 in the existing `if public is True:` above

Issue #291, opened 2019-07-24
Distance matrices biased by sampling depth


clausmith:
This is not a bug in the software, more a concern with the underlying data. Because we don't rarefy, distance matrices calculated using something like `plot_distance` cause otherwise similar samples to cluster by sequencing depth. Take the following set of samples:

<img width="550" alt="Screen Shot 2019-07-24 at 2 29 30 PM" src="https://user-images.githubusercontent.com/2334701/61818733-7617c800-ae1f-11e9-9fd4-1885f4494550.png">

While they all appear quite to have similar compositions, here is the corresponding heatmap of Bray-Curtis distance:

<img width="426" alt="Screenshot 2019-07-24 14 26 39" src="https://user-images.githubusercontent.com/2334701/61818680-57193600-ae1f-11e9-8c8a-9ae34c27a8de.png">

This shows a Bray-Curtis between some of the samples being effectively 1 (i.e. the samples share no taxa in common). This is because Bray-Curtis is being calculated on counts, so samples with a high readcount (e.g. > 1 million) are considered highly dissimilar to samples with a low readcount (< 100k), even if the samples have similar normalized compositions.

I think in a case like this we need to have an option to compute the distance matrix based on normalized counts. Whether we do this through subsampling the counts vectors or multiplying the normalized vectors by a constant to get back to counts is up for discussion. Note, this probably shouldn't be the _default_, but it should at least be available as an option for cases where you need to correct for extreme differences in sequencing depth.

polyatail:
Bray-Curtis distance on normalized read counts is equivalent to 1/2 the manhattan (city block) distance. Isn't that weird? Allowing Bray-Curtis on normalized data is kind of like offering a 0.5 * manhattan option.

We can enable this but I feel like the chances of user error are higher. The example here (I believe) is a legitimate result from Bray-Curtis since there are far more counted organisms in some datasets versus the others, and that is being called out. Rather than being a case of bias I think this is a case of working as intended but not being the correct metric for what we're trying to compare, i.e., the community composition.

Thoughts? Is there a reason to use Bray-Curtis here rather than e.g., UniFrac?

clausmith:
You're right in that this is working as intended, it's just not necessarily obvious what's going on and that the read depths are effectively what's driving the clustering. I think the main issue with UniFrac is that the tree we're using is missing accurate branch lengths, so the UniFrac numbers are all over the place and actually _less_ correct (see #148).

boydgreenfield:
[...]

Issue #285, opened 2019-07-23
Exclude host reads from abundance bar charts


clausmith:
It looks like the `plot_bargraph` helper includes host reads when you plot abundances at the species level:
<img width="962" alt="Screen Shot 2019-07-23 at 3 15 29 PM" src="https://user-images.githubusercontent.com/2334701/61740479-b7936f00-ad5c-11e9-9223-506ed570284f.png">

PR #309, opened 2019-09-05
Changes to viz plotters

polyatail:
This PR addresses:

- mainline/#3229 Changes duplicate labels in plotters, appending an incremented number to the end of each duplicate to avoid collisions in Altair. For example, two samples named `Sample` will become `Sample (1)` and `Sample (2)`
- #285 Drop host reads from `Classifications` results on a per-classification basis. For example, a sample with a human host will lose its human reads, and a sample with a pig host will lose its pig reads.
- Warn user if trying to create a `SampleCollection` containing multiple jobs. For example, a food database job and a One Codex Database job. This should warn users when analyses with multiple different hosts are mixed.
- #291 Enable jaccard, cityblock, and Bray-Curtis beta diversity metrics on normalized read counts. Users must now be careful or they will obtain bad results

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -83,9 +83,6 @@ class AnalysisMixin(
         Multiple metadata fields in a tuple must both be categorical. That is, a numerical field and
         boolean can not be joined, or the result would be something like '87.4_True'.
 
-        The 'Label' field name is transformed to '_display_name'. This lets us label points in plots
-        by the name generated for each sample in `SampleCollection._collate_metadata`.
-
         Returns
         -------
         `pandas.DataFrame`
@@ -149,7 +146,7 @@ class AnalysisMixin(
 
                     if isinstance(label, six.string_types):
                         if label in self.metadata.columns:
-                            magic_metadata[str_f] = self.metadata[label]
+                            magic_metadata[str_f] = self.metadata[label].astype(str)
                         else:
                             raise OneCodexException(
                                 "Label field {} not found. Choose from: {}".format(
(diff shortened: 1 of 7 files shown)


clausmith:
@boydgreenfield @polyatail how do we feel about using abundances by default when available. As an example, when you run the following:

```
from onecodex import Api
from onecodex.models import SampleCollection

ocx = Api()

project = ocx.Projects.where(name="<project name>")[0]

samples = ocx.Samples.where(project=project)

samples.to_df()
```

You get a `dataframe` containing normalized `readcount_w_children`, even those these samples all have `abundances` calculated for them. It might be nice to default to using `abundances` as the `field` when available. Thoughts?

boydgreenfield:
> @boydgreenfield @polyatail how do we feel about using abundances by default when available. As an example, when you run the following:

I like this, but it's something of a breaking change. Maybe do it and bump to `0.7.0`?

polyatail:
I _kind of_ feel like not always returning the same thing could be confusing for users. For example, let's say Samples 1-10 all have abundances, and so the resulting df displays abundances. Then, you add Sample 11, which has only readcount_w_children. Now, the resulting df displays only readcount_w_children since abundances are not available for all samples. Suddenly, all the results for Samples 1-10 have changed, and you're all like wth?

clausmith:
Understood, and good point. I think if we take a step back, the priorities are the following:

1. Provide a good user experience for novice users (I.e. minimal foot guns)

2. Default to using the _best data available_ at all times, unless it contradicts 1.

With that in mind, our abundances really are better data, and they’re what we present by default in the interface when available. 

I propose whenever someone constructs a sample collection that has some but not all samples with abundances and does not pass `field` explicitly, we print a warning saying we are falling back to `readcount_w_children` since not all samples have abundances. Otherwise if all samples have abundances, we proceed with abundances.

PR #321, opened 2019-10-14
Create a session to upload samples/documents

Keats:
Attempt to avoid some thread safety issues with S3 client as we are getting some `KeyError: 'endpoint_resolver'` which might be due to the way we were instantiating the client.

The boto3 docs recommend to use a session for each thread: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-multiprocessing
The example uses a resource instead of a client but I am assuming it's the same thing?


boydgreenfield:
@Keats This looks good to me and is indeed the correct usage. Merging now and I'm going to do some reliability testing later before pinning a release.

PR #325, opened 2019-12-03
Adds option to plot 'Other' bars on taxonomic bar charts

### LINKED ISSUE #288: Add Other taxa to plot_bargraph
The `plot_bargraph` helper should include an additional "Other" row that is basically equal to `1 - sum(top_n_abundances)`. Extra points if we can make it always coloured grey.

clausmith:
This PR adds a kwarg to the `plot_bargraph` function that lets the user plot the remaining taxa as an "Other" bar in the taxonomic bar chart. It always plots "Other" as a grey bar at the bottom of the graph, and supports both normalized and absolute readcounts.

Closes #288.

### DIFF
--- a/onecodex/viz/_bargraph.py
+++ b/onecodex/viz/_bargraph.py
@@ -151,12 +151,6 @@ class VizBargraphMixin(object):
         dfs_to_plot = []
 
         if haxis:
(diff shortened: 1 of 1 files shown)


boydgreenfield:
Working on this separately. Closing for now.

PR #331, opened 2020-03-12
Update altair to 4.0

Keats:
So we can have the nice new features.
Altair v4 drops support for python2 so that's a breaking change


Keats:
Should we remove python2 support entirely? It's past end of life now

Keats:
I think I would recommend dropping python2 support from master but keep a branch for 0.7 where we can commit fixes for py2 if needed

boydgreenfield:
Closing in favor of #338.

PR #338, opened 2020-03-25
Altair 4 update and report improvements

### LINKED ISSUE #333: Automatically set ONE_CODEX_REPORT_UUID in report footers when in environment

boydgreenfield:
- Updates to use `altair_saver` for generation of HTML figures and SVG images inline in our notebooks
- Sets the report UUID if `ONE_CODEX_REPORT_UUID` environment variable is speficified. Closes #333.
- Updates `setup.py` to have a `[reports]` `extras_requires`. This makes it _easier_ to install all the report dependencies into a simple notebook. Probably more work to further ease that (including more documentation). See https://github.com/onecodex/sars-cov-2/blob/master/Dockerfile for an example.


Keats:
Anything blocking the merge/release?

onecodex/notebooks/exporters.py
```diff
@@ -190,6 +168,10 @@ class OneCodexPDFExporter(OneCodexHTMLExporter):
     export_from_notebook = "One Codex PDF Report"
     template_path = [ASSETS_PATH]
 
+    def __init__(self, config=None, **kw):
+        super(OneCodexPDFExporter, self).__init__(config=config, **kw)
```
clausmith, line 172: Since we're just supporting Python 3+ here, I think this could be simplified to just `super().__init__(config=config, **kw)`. They're basically the same, just a matter of brevity.

boydgreenfield: I'm trying to keep this code a bit Python 2-y so we don't accidentally break stuff that _might_ work on Python 2, but fair point. I think I'll leave this as you _could_ use the report features technically in Python 2 I think (i.e., if you got all the deps installed and didn't need our Pandas integrations that are Python 3.4+ only).

polyatail:
Awesome, that `altair_saver` packages is just what we needed!

PR #348, opened 2020-04-09
Update vendored potion-client library for compatibility with collecti…

boydgreenfield:
…ons imports in Python 3.9


Keats:
Looks like we need to remove circle from the repo, I can't merge it otherwise.

boydgreenfield:
I removed it on the CircleCI side... but there’s something funky with the way their GitHub app is behaving. I’ll merge this and try to disable the checks again. :(

boydgreenfield:
@Keats Turns out it's because I didn't disable it in the branch protection rules. 😬 Fixed now.

PR #350, opened 2020-04-29
Plotting and data analysis interface improvements

### LINKED ISSUE #288: Add Other taxa to plot_bargraph

### LINKED ISSUE #305: Allow simple resizing of plots
It would be nice to have `width` and `height` kwargs that let you specify the size of the output. These can just be passed directly to Altair's `.properties()` function I believe, but we'd want to make sure that whatever is passed to the helper is the _total_ plot size, including legends and other features.

One quick gotcha here is since the heatmaps returned by `plot_distance` are `HConcat` chart objects, you can't just call `.properties()` on the resulting object.

### LINKED ISSUE #306: Add support for calculating alpha diversities on abundances
Right now, we only support calculating alpha and beta diversities on the unnormalized read counts. For data with estimated abundances, we should allow calculating at least alpha diversity based on these abundances. 

Unfortunately Bray-Curtis only works with counts data, not proportions.

### LINKED ISSUE #310: Use abundances by default
When a new `SampleCollection` is instantiated, try to use abundance data if it is available for all samples. If it is not, issue a warning and fallback to readcount_w_children.

### LINKED ISSUE #345: Add support for calculating Unifrac distance matrices
In our `beta_diversity()` functions, we only support `braycurtis`, `jaccard`, and `cityblock` as values for the `metric` arg. 

We already support calculating Unifrac (weighted and unweighted) distances in `plot_distance`, so we should add support for outputting these as distance matrices as well.

clausmith:
This PR makes a handful of changes with the following two objectives:

1. Improve user ergonomics of plotting and distance functions
2. Improving our defaults around which metrics, ranks, and other parameters to use when

The main changes in here are:

- Adds more pythonic support for truthy/falsey values to `SampleCollection.filter()`
- Adds support for passing sorted lists to the various `sort_x` and `sort_y` plotting arguments

- Adds a property to `SampleCollection` to track whether we think the collection is all WGS/metagenomic data
- Checks the above and defaults to using `abundances` data for metagenomic datasets instead of `readcount_w_children`
- `rank="auto"` now defaults to `species` if the `field="abundances"` or the dataset is metagenomic, instead of `genus`
- Adds support for specifying `weighted_unifrac` and `unweighted_unifrac` as metrics in the `SampleCollection.beta_diversity()` method
- Adds support for calculating alpha and beta diversities on normalized data*. This is important since we'll now be using abundances by default for _most_ datasets
- Changes default alpha diversity metric from `simpson` to `shannon`, since `shannon` is generally a more appropriate default
- Adds support for specifying chart `width` and `height` directly in the `plot_*` functions
- Moves to use native Altair box plots and native Altair faceting (yay for simplification!)
- Adds option to plot "Other" bars on bargraphs with normalized read counts and abundances
- Adds option to plot "No <rank>" bars on bargraphs with abundances
- Removes support for just specifying `unifrac` as a metric - explicit over implicit
- Adds abundance rollups so we can use abundances at all taxonomic ranks
- Enumifies many things
- Switches to using a normalized tree for Weighted Unifrac calculations, which gives us Unifrac values in a `[0, 1]` range
- Adds a `project` column to the `metadata` DataFrame of a `SampleCollection`

\* The `scikit-bio` docs specify that the inputs to the alpha diversity functions should be counts. But if you look at the source, they proportionalize the data right away _anyway_, so I feel comfortable just passing in proportions (such as abundances) directly. For discussion of using normalized data with beta diversity functions, see #291. 

Closes #345 
Closes #306 
Closes #310 
Closes #288 
Closes #305


onecodex/analyses.py
```diff
@@ -49,7 +49,7 @@ def _guess_normalized(self):
         """
         return (
             getattr(self, "_normalized", False)
-            or getattr(self, "_field", None) == "abundance"
+            or getattr(self, "_field", None) in ["abundance", "abundance_w_children"]
```
Keats, line 52: can we enumify all those abundance/abundance_w_children and the ranks?

tests/test_dataframes.py
```diff
@@ -64,3 +65,30 @@ def test_pandas_extension(ocx, api_data):
         "204441",
         "1783270",
     ]
+
+# Current, there's a bug in pandas where these types of operations don't carryover
```
Keats, line 69: do you a have a link to a GH issue for it?

onecodex/analyses.py
Keats: Nevermind, just saw it's on your todo list

onecodex/models/collection.py
```diff
@@ -274,6 +289,25 @@ def _collate_results(self, field=None, include_host=None):
             table = results["table"]
             host_tax_ids = results.get("host_tax_ids", [])
 
+            table = {t.get("tax_id"): t for t in table}
```
Keats, line 292: Can there be cases where `t` doesn't have tax_id? Just a bit below it's doing `["tax_id"]`

onecodex/models/collection.py
```diff
@@ -274,6 +289,25 @@ def _collate_results(self, field=None, include_host=None):
...
+                result["abundance_w_children"] = result["abundance"]
+
+                while parent:
+                    if "abundance_w_children" not in parent:
+                        parent["abundance_w_children"] = 0
+
+                    parent["abundance_w_children"] += result["abundance"]
+                    parent = table.get(parent["parent_tax_id"])
+
+            table = [v for k, v in table.items()]
```
Keats, line 309: `table.values()`? 
Can we have a different name for the dict version since it doesn't have the same type?

onecodex/viz/_bargraph.py
```diff
@@ -134,82 +154,65 @@ def plot_bargraph(
         # OCX this should be okay)
         #
 
-        ylabel = df.ocx.field if ylabel is None else ylabel
+        ylabel = field if ylabel is None else ylabel
```
Keats, line 157: `ylabel or field`?

onecodex/viz/_bargraph.py
```diff
@@ -134,82 +154,65 @@ def plot_bargraph(
...
         xlabel = "" if xlabel is None else xlabel
```
Keats, line 158: xlabel or "" to match my comment above.

onecodex/viz/_bargraph.py
```diff
@@ -134,82 +154,65 @@ def plot_bargraph(
...
-                    ),
-                    color=alt.Color("tax_name", legend=alt.Legend(title=legend)),
-                    tooltip=tooltip_for_altair,
-                    href="url:N",
-                )
+            kwargs["column"] = haxis
+
+        domain = sorted(df["tax_name"].unique())
+
+        no_level_name = "No {}".format(rank)
```
Keats, line 172: I think we can generate all of those at the module level in a dict? Not that this would make any differences for performance but maybe a bit cleaner.

tests/test_dataframes.py
boydgreenfield: I'm not sure if @clausmith has opened it yet. @clausmith please add once open on https://github.com/pandas-dev/pandas

onecodex/lib/enums.py
```diff
@@ -0,0 +1,13 @@
...
+
+BETA_DIVERSITY_METRICS = {
+    "jaccard",
+    "braycurtis",
+    "cityblock",
+    "weighted_unifrac",
+    "unweighted_unifrac",
+}
+
+FIELDS = {"auto", "abundance", "readcount", "readcount_w_children"}
```
Keats, line 13: Can we make all of those enums? We compare to a string everywhere in the codebase.

tests/test_dataframes.py
Keats: can we create it before merging?

clausmith: Yes I think so. There are about half a dozen other issues open already about this functionality, and there's some discussion about deprecating it completely. I'm just trying to figure out whether there's already an issue open specifically about _our_ bug.

onecodex/analyses.py
```diff
@@ -303,10 +304,9 @@ def to_df(
         if normalize is False and self._guess_normalized():
             raise OneCodexException("Data has already been normalized and this can not be undone.")
 
-        if normalize is True or (
-            normalize == "auto" and rank is not None and self._field != "abundance"
-        ):
-            df = df.div(df.sum(axis=1), axis=0)
+        if normalize is True or (normalize == "auto" and rank):
+            if not self._guess_normalized():
+                df = df.div(df.sum(axis=1), axis=0)
```
clausmith, line 309: @boydgreenfield I figured out what's going on here. We've introduced `abundance_w_children` as a field for cumulative abundances, which opened a bug where you could double-normalize the abundance data. I moved the check for whether the data type is abundances into `_guess_normalized`.

onecodex/analyses.py
```diff
@@ -357,7 +357,7 @@ def _make_pretty_field_name(field, normalized):
                 return "Reads (Normalized)"
             else:
                 return "Reads"
-        elif field == "abundance":
+        elif field == Field.Abundance:
             return "Relative Abundance"
```
boydgreenfield, line 361: We should probably use the `Field` enum in line 355 too

onecodex/distance.py
```diff
@@ -21,28 +22,20 @@ def alpha_diversity(self, metric="simpson", rank="auto"):
         """
         import skbio.diversity
 
-        if metric not in ("simpson", "chao1", "shannon"):
+        if metric not in AlphaDiversityMetric.values():
```
boydgreenfield, line 25: ~I kind of dislike this. @Keats what do you think about a `has_value` staticmethod or similar? That seems to be a suggestion here and is IMO a bit clearer than adding `values` in this same way.~

@clausmith @Keats already added this. I'd just lean towards writing as `AlphaDiversityMetric.has_value(metric)`

onecodex/lib/enums.py
```diff
@@ -0,0 +1,74 @@
...
+    def values(cls):
+        return [e.value for e in cls]
+
+
+class AbundanceField(BaseEnum):
+    Abundance = "abundance"
+    AbundanceWChildren = "abundance_w_children"
+
+
+class Field(BaseEnum):
```
boydgreenfield, line 21: Should this be renamed to `Metric` and `AbundanceMetric`. "Field" feels very generic. Realize that's a larger API change probably...

onecodex/models/collection.py
```diff
@@ -140,42 +151,42 @@ def _classification_fetch(self, skip_missing=None):
...
 
         new_classifications = []
 
-        for a in self._res_list:
-            if a.__class__.__name__ == "Samples":
-                c = a.primary_classification
-            elif a.__class__.__name__ == "Classifications":
-                c = a
+        for obj in self._res_list:
+            if isinstance(obj, Samples):
```
boydgreenfield, line 161: 👏 Thank you.

onecodex/models/collection.py
```diff
@@ -255,7 +266,7 @@ def _collate_results(self, field=None, include_host=None):
         field = field if field else self._kwargs["field"]
         include_host = include_host if include_host else self._kwargs["include_host"]
 
-        if field not in ("auto", "abundance", "readcount", "readcount_w_children"):
+        if field not in Field.values():
```
boydgreenfield, line 269: Nit: `Field.has_value(field)`, but see my point re: field vs. metric as a name here.

onecodex/lib/enums.py
Keats: @clausmith I've updated the has_value calls, i'll let you weight in on the naming. Let me know what you want/prefer and I'll update it.

clausmith: There are one or two other places where we have a kwarg for `field`, but I don't have a strong preference either way. As long as `Metric` wouldn't get confusing with `AlphaDiversityMetric` and `BetaDiversityMetric` since it's distinct then I think we're fine to change `Field` -> `Metric`.

onecodex/models/collection.py
clausmith: You did this.

onecodex/analyses.py
```diff
@@ -303,10 +304,9 @@ def to_df(
         if normalize is False and self._guess_normalized():
             raise OneCodexException("Data has already been normalized and this can not be undone.")
 
-        if normalize is True or (
-            normalize == "auto" and rank is not None and self._field != "abundance"
-        ):
-            df = df.div(df.sum(axis=1), axis=0)
+        if normalize is True or (normalize == "auto" and rank):
```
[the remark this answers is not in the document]
clausmith: The abundance check is moved to `_guess_normalized`

CHANGELOG.md
```diff
@@ -0,0 +1,31 @@
...
+-   Switches to using a normalized tree for Weighted Unifrac calculations, which gives us Unifrac values in a [0, 1] range
+-   Defaults to using abundances data for metagenomic datasets instead of readcount_w_children
+
+### Deprecated
+
+-   The `field` kwarg on the `SampleCollection` constructor has been renamed to `metric`. We still support passing either, but show a `DeprecationWarning` when passing `field`
+
+### Removed
+
+-   Removes support for just specifying `unifrac` as a metric
```
Keats, line 31: is there a recommended new value for people using that metric? unweighted_unifrac?

onecodex/analyses.py
```diff
@@ -49,7 +50,7 @@ def _guess_normalized(self):
         """
         return (
             getattr(self, "_normalized", False)
-            or getattr(self, "_field", None) == "abundance"
+            or AbundanceMetric.has_value(self._metric)
```
Keats, line 53: so `self._metric` cannot be None/undefined anymore?

onecodex/distance.py
```diff
@@ -58,23 +51,28 @@ def beta_diversity(self, metric="braycurtis", rank="auto"):
...
 
-        counts = []
-        for c_id in df.index:
-            counts.append(df.loc[c_id].tolist())
+        if metric == BetaDiversityMetric.WeightedUnifrac:
+            return self.unifrac(weighted=True, rank=rank)
+        elif metric == BetaDiversityMetric.UnweightedUnifrac:
+            return self.unifrac(weighted=False, rank=rank)
+        elif metric == BetaDiversityMetric.Jaccard:
+            df = df > 0  # Jaccard requires a boolean matrix, otherwise it throws a warning
```
Keats, line 68: `df > 0`... what does it even do? replace by a boolean every value that is > 0 in the dataframe?

onecodex/models/analysis.py
```diff
@@ -40,6 +51,51 @@ class Classifications(Analyses):
...
+        # Roll-up abundances to parent taxa
+        for tax_id, result in table.items():
+            if result["parent_tax_id"] not in table:
+                continue
+
+            parent = table[result["parent_tax_id"]]
+            result[Metric.AbundanceWChildren.value] += result[Metric.Abundance] or 0.0
+
+            while parent:
+                parent[Metric.AbundanceWChildren.value] += result[Metric.Abundance] or 0.0
```
Keats, line 87: I'm not sure you need all the `.value` in that function

onecodex/viz/_bargraph.py
```diff
@@ -134,82 +152,93 @@ def plot_bargraph(
...
+            "#76b7b2",
+            "#59a14f",
+            "#edc948",
+            "#b07aa1",
+            "#ff9da7",
+            "#9c755f",
+            "#bab0ac",
+        ]
+        other_colour = ["#d0dadb"]
+        no_level_colour = ["#eeefe1"]
```
Keats, line 183: Can we extract the colours outside of the function or we never want to override/change it?

onecodex/viz/_bargraph.py
```diff
@@ -134,82 +152,93 @@ def plot_bargraph(
...
+
+        chart = (
+            alt.Chart(df)
+            .mark_bar()
+            .encode(
+                x=alt.X("Label", axis=alt.Axis(title=xlabel), sort=sort_order),
+                y=alt.Y(
+                    self.metric, axis=alt.Axis(title=ylabel), scale=alt.Scale(**y_scale_kwargs)
+                ),
+                color=alt.Color(
```
Keats, line 216: sadly it uses the wrong spelling here :(

onecodex/viz/_distance.py
```diff
@@ -14,25 +15,26 @@ def _compute_distance(self, rank, metric):
...
+        elif metric in (BetaDiversityMetric.BrayCurtis, "bray-curtis", "bray curtis"):
+            distances = self.beta_diversity(metric=BetaDiversityMetric.BrayCurtis, rank=rank)
+        elif metric in ("manhattan", BetaDiversityMetric.CityBlock):
+            distances = self.beta_diversity(metric=BetaDiversityMetric.CityBlock, rank=rank)
+        elif metric == BetaDiversityMetric.Jaccard:
+            distances = self.beta_diversity(metric=BetaDiversityMetric.Jaccard, rank=rank)
+        elif metric == BetaDiversityMetric.WeightedUnifrac:
             distances = self.unifrac(weighted=True, rank=rank)
-        elif metric == "unweighted_unifrac":
+        elif metric == BetaDiversityMetric.UnweightedUnifrac:
```
Keats, line 26: If we expose the enums and recommend their usage in the docs, we can just let people use that and not have to worry about mispelling/other names and they would get auto-completion in notebooks I think?

onecodex/viz/_metadata.py
```diff
@@ -168,20 +167,31 @@ def plot_metadata(
...
+
+        if title:
+            props["title"] = title
+        if width:
+            props["width"] = width
+        if height:
+            props["height"] = height
+
+        if props:
+            chart = chart.properties(**props)
```
Keats, line 193: this code is repeated a few times, something updating the chart title too. Can it be abstracted? Not a blocker

onecodex/viz/_primitives.py
```diff
@@ -4,72 +4,21 @@
...
-    upper_plot = alt.Chart(df).mark_rule().encode(y=upper_whisker, y2=upper_box, x=x_format)
-
-    middle_tick = (
-        alt.Chart(df)
-        .mark_tick(color="black", size=35)
-        .encode(
-            y="median({}):Q".format(quantity),
-            x=alt.X(x_format, axis=alt.Axis(title=xlabel)),
-            tooltip="median({}):Q".format(quantity),
+def sort_helper(sort, values):
```
Keats, line 7: Can this have a docstring? Especially when `sort` is a list, it seems to be a `set` version of `values`?

onecodex/viz/_primitives.py
```diff
@@ -4,72 +4,21 @@
...
+    sort_order = None
+    values = list(set(values))
```
Keats, line 9: you could delay that in the branch below as it is only in that case that you need to do it.

CHANGELOG.md
clausmith: Generally speaking the user should specify whether they want to do `unweighted_unifrac` or `weighted_unifrac`. Just specifying `unifrac` was vague, and didn't really match up with the usual terminology.

onecodex/analyses.py
clausmith: No, it shouldn't be.

onecodex/distance.py
clausmith: Yep, exactly. `jaccard` throws a warning if you pass it a quantitative data frame, since it then does the conversion implicitly.

onecodex/viz/_bargraph.py
clausmith: This colour scale is specific to the bar plot here. We can look at moving it out in the future if we have other plots that use the bar plot format.

onecodex/viz/_bargraph.py
clausmith: I know :(

onecodex/viz/_distance.py
clausmith: I think our users are probably at the level where enums are pretty far outside their scope of comprehension. I'd prefer to let them just pass in strings for now.

onecodex/models/analysis.py
boydgreenfield: We do need it or the key becomes the enum, which is a bit gross as it's inconsistent with the other keys.

onecodex/viz/_distance.py
boydgreenfield: Agree with @clausmith.

onecodex/viz/_metadata.py
boydgreenfield: Simplified.

onecodex/viz/_primitives.py
boydgreenfield: Updated.

onecodex/viz/_primitives.py
boydgreenfield: 👍

Issue #248, opened 2019-04-14
Speed up library load time for CLI operation


boydgreenfield:
Currently, we load a bunch of 3rd party libraries _when the user installs with `pip install onecodex[all]`_, which can add significant interpreter startup overhead to the CLI. We should profile and reduce this overhead where possible.

boydgreenfield:
See this article for some profiling advice in Python 3.7: https://dev.to/methane/how-to-speed-up-python-application-startup-time-nkf

boydgreenfield:
With Python 3.7 installed: `python -X importtime -c 'import onecodex' 2> onecodex-imports`

jairideout:
Hi there! I'm interested in working on this issue if someone isn't already tackling it. Is it okay if I submit a pull request?

boydgreenfield:
@jairideout Absolutely! Though FYI we did make a dent in this ~9 months ago and maybe should have closed this one (see https://github.com/onecodex/onecodex/blob/master/tests/test_speed.py). But if you can make it a bit faster (or if we've had regressions 😬), 100% that'd be fantastic.

jairideout:
Great, I'll take a look!

PR #352, opened 2020-06-01
Faster imports for `viz` and `analyses` subpackages

jairideout:
## Description

Sped up import times for `onecodex.viz` and `onecodex.analyses` subpackages. Testing locally on a macOS laptop, the previous import time for each package was ~670ms, and the new import time is ~170ms.

Faster imports are achieved by defining an import hook to configure Altair's theme and renderer the first time it is imported. Directly importing and configuring Altair in the `onecodex.viz` subpackage was the primary cause of the slow imports. An import hook avoids this performance hit by configuring Altair during deferred import in visualization code.

I also deferred a few more imports in the `viz` subpackage, and added a couple of unit tests to ensure Altair is being configured as expected.

Addresses part of #248.

## Question for reviewers

The import hook code is Python 2/3 compatible by using the `imp` package, which is deprecated in Python 3. Do we want to use the current implementation, or is it preferable to drop support for Python 2 and use the `importlib` package?

### DIFF
--- a/onecodex/distance.py
+++ b/onecodex/distance.py
@@ -1,5 +1,3 @@
-import pandas as pd
-
 from onecodex.exceptions import OneCodexException
 from onecodex.taxonomy import TaxonomyMixin
 
@@ -19,6 +17,7 @@ class DistanceMixin(TaxonomyMixin):
         -------
         pandas.DataFrame, a distance matrix.
         """
+        import pandas as pd
         import skbio.diversity
 
         if metric not in ("simpson", "chao1", "shannon"):
--- a/onecodex/models/__init__.py
+++ b/onecodex/models/__init__.py
@@ -372,7 +372,7 @@ class OneCodexBase(object):
         # FIXME: this doesn't actually work though, because potion creates these routes for all
         # items :/
         method_links = cls._resource._schema["links"]
-        return any(True for l in method_links if l["rel"] == method_name)
+        return any(True for link in method_links if link["rel"] == method_name)
 
     @classmethod
     def all(cls, sort=None, limit=None):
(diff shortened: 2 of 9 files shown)


jairideout:
I think the pre-commit `flake8` errors are unrelated to my changes (I'm seeing them on the current `master` branch too). Would you like me to fix them here in a second commit, or is a separate PR preferred?

boydgreenfield:
Awesome, thanks @jairideout!

> The import hook code is Python 2/3 compatible by using the imp package, which is deprecated in Python 3. Do we want to use the current implementation, or is it preferable to drop support for Python 2 and use the importlib package?

It's fine to use `importlib`. We support Python 2 _only for the CLI_, the remainder of the library requires Python 3.4+ (and later versions of Pandas do not support Python 2 either).

> I think the pre-commit `flake8` errors are unrelated to my changes (I'm seeing them on the current `master` branch too). Would you like me to fix them here in a second commit, or is a separate PR preferred?

A separate commit in this PR is fine.

boydgreenfield:
@jairideout This now has a merge conflict with changes I just pushed on master. I'd like to:
- [ ] Resolve that merge conflict
- [ ] Update to use `importlib` since we don't support Python 2 for the visualization functions _anyway_

If you're OK with it (or I don't hear by later this evening), I'll probably just do that in a commit and then merge your changes in. We're about to pin a `v0.8.0` release, so I'd love to get the speedups in for that! 👏

jairideout:
Thanks for the heads up, @boydgreenfield! I'm resolving the merge conflicts now. It looks like the pre-commit `flake8` errors were fixed on `master` (thanks!), and I'll need to revert my [separate commit](https://github.com/onecodex/onecodex/pull/352/commits/891ab76e9d75976307eeabd63fe24b8d8b18021a) since it's no longer necessary. I'm going to close this PR and open a new one to keep the commit history clean.

jairideout:
The new PR is here: #353

PR #353, opened 2020-06-02
Faster imports for `viz` and `analyses` subpackages

jairideout:
## Description

This PR replaces #352.

Sped up import times for `onecodex.viz` and `onecodex.analyses` subpackages. Testing locally on a macOS laptop, the previous import times for both packages were ~670ms, and the new import times are ~170ms.

Faster imports are achieved by defining an import hook to configure Altair's theme and renderer the first time it is imported. Directly importing and configuring Altair in the `onecodex.viz` subpackage was the primary cause of the slow imports. An import hook avoids this performance hit by configuring Altair during deferred import in visualization code.

I also deferred a few more imports in the `viz` subpackage, and added a couple of unit tests to ensure Altair is being configured as expected.

Addresses part of #248.

## TODO

The import hook code is Python 2/3 compatible by using the `imp` package, which is deprecated in Python 3. I spent some time today attempting to update the code to use Python 3's `importlib` package but wasn't able to get it to work. I'm happy to continue working on this but likely won't have it ready for the `v0.8.0` release today. @boydgreenfield, please feel free to push additional commits if you have time to work on the `importlib` issue in the meantime. Thanks!

### DIFF
--- a/onecodex/distance.py
+++ b/onecodex/distance.py
@@ -1,5 +1,3 @@
-import pandas as pd
-
 from onecodex.exceptions import OneCodexException
 from onecodex.taxonomy import TaxonomyMixin
 from onecodex.lib.enums import AlphaDiversityMetric, BetaDiversityMetric, Rank
@@ -20,6 +18,7 @@ class DistanceMixin(TaxonomyMixin):
         -------
         pandas.DataFrame, a distance matrix.
         """
+        import pandas as pd
         import skbio.diversity
 
         if not AlphaDiversityMetric.has_value(metric):
--- a/onecodex/viz/__init__.py
+++ b/onecodex/viz/__init__.py
@@ -1,4 +1,5 @@
-import altair as alt
+import imp
+import sys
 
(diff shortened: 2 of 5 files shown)


boydgreenfield:
@jairideout No worries – if `imp` isn't throwing deprecation warnings into the console, it's probably fine to just use it for now too. I'll take a peek and merge if so; otherwise I'll pin 0.8.0 and we'll figure it out.

jairideout:
Sounds good! I'm not seeing deprecation warnings when importing `onecodex.viz`, `onecodex.analyses`, or `altair` in an interpreter because Python ignores `DeprecationWarning` by default (unless it is raised in `__main__`). Thanks for taking a look!

boydgreenfield:
@jairideout tested and this looks great. Sneaking it into our v0.8.0 cut. Thanks again!

jairideout:
@boydgreenfield thanks for testing and including it in the `v0.8.0` release! 🎉

PR #354, opened 2020-06-05
Minor bugfixes for a v0.8.1 release

clausmith:
This PR includes 2 principal bug fixes/enhancements:

1. Adds better retry logic for the upload confirmation request. If this request failed for any reason, it cancels the rest of the uploads. Now it retries with a conservative backoff.

2. Fixes a bug with the "Export to One Codex" upload flow.

The objective is to push a `v0.8.1` release. 

Fixes DEV-3915, fixes DEV-3921, fixes DEV-2703.

### DIFF
--- a/onecodex/lib/upload.py
+++ b/onecodex/lib/upload.py
@@ -9,6 +9,7 @@ import six
 from onecodex.exceptions import OneCodexException, raise_connectivity_error, raise_api_error
 from onecodex.utils import atexit_register, atexit_unregister, snake_case
 from onecodex.lib.files import FilePassthru, get_file_wrapper
+from urllib3.util.retry import Retry
 
 
 log = logging.getLogger("onecodex")
@@ -354,8 +355,6 @@ def _upload_document_fileobj(file_obj, file_name, documents_resource):
         `FilePassthru` will send a compressed type if the file is gzip'd or bzip'd.
     file_name : `string`
         The file_name you wish to associate this file with at One Codex.
-    fields : `dict`
-        Additional data fields to include as JSON in the POST.
     documents_resource : `onecodex.models.Documents`
         Wrapped potion-client object exposing `init_upload` and `confirm_upload` routes to mainline.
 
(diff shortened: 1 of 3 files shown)


onecodex/notebooks/exporters.py
```diff
@@ -209,12 +210,14 @@ def from_notebook_node(self, nb, resources=None, **kw):
...
             document_id = _upload_document_fileobj(
-                BytesIO(output), file_name, ocx._client.session, ocx.Documents._resource
+                BytesIO(output), file_name, ocx.Documents._resource
             )
         except UploadException as exc:
             resp = json.dumps({"status": 500, "message": str(exc)})
             return resp, resources
-        except Exception:
+        except Exception as exc:
+            client = get_raven_client()
```
boydgreenfield, line 219: Do we need to do this anymore? I _think_ in the newest versions of the Sentry library you can just do `sentry_sdk.capture_exception` but maybe we need to upgrade.

onecodex/lib/upload.py
```diff
@@ -482,13 +481,15 @@ def _s3_intermediate_upload(file_obj, file_name, fields, session, callback_url):
...
     try:
+        # retry on 502, 503, 429, with a backoff timing of 4s, 8s, and 16s
         resp = session.post(
             callback_url,
             json={
                 "s3_path": "s3://{}/{}".format(fields["s3_bucket"], fields["file_id"]),
                 "filename": file_name,  #
                 "import_as_document": fields.get("import_as_document", False),
             },
+            max_retries=Retry(total=3, backoff_factor=4, status_forcelist=[502, 503, 429]),
```
boydgreenfield, line 492: I believe `method_whitelist` needs to explicitly be specified as POSTs aren't considered idempotent. We _probably_ want to retry on a generic 500 too, though that really shouldn't be happening (in general, I'd avoid this as we don't know it's idempotent, but in this case the server-side operation _is_).

boydgreenfield:
Closes DEV-3915, DEV-3921, DEV-2703.

PR #358, opened 2020-07-08
Improve input validation in plotting methods

jairideout:
Improved input validation for the various `plot_*` methods in the `onecodex.viz` subpackage:

- Added `onecodex.exceptions.PlottingException` for reporting user-facing plotting errors (e.g. for the plot builder). This exception type is a subclass of `OneCodexException`.

- Added `onecodex.exceptions.PlottingWarning` for reporting user-facing plotting warnings (e.g. for the plot builder). This warning type is a subclass of `UserWarning`.

- Methods now require that the minimum number of samples and/or taxa are present for a given plot type. This number varies by type, and is usally a minimum of 1, 2, or 3 samples or taxa.

- `plot_metadata` now raises a `PlottingWarning` if any boxplot groups of size 1 are detected (groups of size 1 are not displayed by Altair, and the plot builder may wish to catch and display this warning to users).

@clausmith and @lynchde, please let me know if there are any additional checks you'd like me to add. Thanks!

### DIFF
--- a/onecodex/viz/_bargraph.py
+++ b/onecodex/viz/_bargraph.py
@@ -1,4 +1,4 @@
-from onecodex.exceptions import OneCodexException
+from onecodex.exceptions import OneCodexException, PlottingException
 from onecodex.lib.enums import AbundanceMetric, Rank, Metric
 from onecodex.viz._primitives import sort_helper, prepare_props
 
@@ -80,6 +80,12 @@ class VizBargraphMixin(object):
 
         if not (threshold or top_n):
             raise OneCodexException("Please specify at least one of: threshold, top_n")
+
+        if len(self._results) < 1:
+            raise PlottingException(
+                "There are too few samples for bargraph plots after filtering. Please select 1 or "
+                "more samples to plot."
+            )
 
         if top_n == "auto" and threshold == "auto":
             top_n = 10
--- a/onecodex/viz/_distance.py
+++ b/onecodex/viz/_distance.py
@@ -3,7 +3,7 @@ from itertools import chain
 import warnings
 
 from onecodex.lib.enums import BetaDiversityMetric, Rank, Linkage, OrdinationMethod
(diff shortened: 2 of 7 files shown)


onecodex/viz/_distance.py
```diff
@@ -289,8 +290,11 @@ def plot_mds(
         from sklearn import manifold
         from sklearn.metrics.pairwise import euclidean_distances
 
-        if len(self._results) < 2:
-            raise OneCodexException("`plot_mds` requires 2 or more valid classification results.")
+        if len(self._results) < 3:
```
petraszd, line 293: Very small nitpick: If you would put `3` to constant then you can re-use it in error's message:

```python
MIN_SAMPLES_REQUIRED = 3  # Needs better name
if len(self._results) < MIN_SAMPLES_REQUIRED:
    raise PlottingException(f"... select {MIN_SAMPLES_REQUIRED} or more ...")
```

Feel free to ignore this comment. It really does not need to be changed.

petraszd:
Kudos for tests!

onecodex/viz/_bargraph.py
```diff
@@ -81,6 +81,12 @@ def plot_bargraph(
         if not (threshold or top_n):
             raise OneCodexException("Please specify at least one of: threshold, top_n")
 
+        if len(self._results) < 1:
```
clausmith, line 84: Let's change this to be after line 98 (`df = self.to_df(...)`) so that we catch instances where there are no samples left after we filter for taxa.

jairideout: I couldn't get `to_df()` to drop samples during taxa filtering. Even if all taxa are filtered out, the returned `df` contains the input samples.

Also, I found that `to_df()` raises a `OneCodexException` if the sample collection is empty, so I don't think we can put the check after `to_df()` because it won't be reached before the other error is raised.

Do you think it makes sense to leave this check where it's at?

PR #359, opened 2020-07-14
Fixes bug where scikit-bio would fail when plotting Weighted Unifrac distances

clausmith:
There was a bug introduced in `v0.8.0` where I had removed some code that was required to workaround a limitation in `scikit-bio`. This PR re-implements that workaround.

### DIFF
--- a/onecodex/distance.py
+++ b/onecodex/distance.py
@@ -121,11 +121,11 @@ class DistanceMixin(TaxonomyMixin):
                 BetaDiversityMetric.WeightedUnifrac,
                 df,
                 df.index,
-                tree=tree,
+                tree=new_tree,
                 otu_ids=tax_ids,
                 normalized=True,
             )
         else:
             return skbio.diversity.beta_diversity(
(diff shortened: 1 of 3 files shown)


onecodex/distance.py
```diff
@@ -102,6 +102,19 @@ def unifrac(self, weighted=True, rank=Rank.Auto):
...
+        #
+        # https://github.com/biocore/scikit-bio/blob/f3ae1dcfe8ea88e52e19f6693d79e529d05bda04/skbio/diversity/_util.py#L89
+        #
+        # Our taxonomy root regularly has more than 2 children, so we
+        # add a fake parent of `root` to the tree here.
+        from skbio.tree import TreeNode
+
+        new_tree = TreeNode(name="fake root")
+        new_tree.rank = "no rank"
+        new_tree.append(tree)
```
boydgreenfield, line 116: Should `new_tree` be passed into the `beta_diversity` call below? And can we add a test for this?

clausmith: Resolved in 963de54

PR #360, opened 2020-07-24
Replacing chao1 with observed_taxa

lynchde:
Chao1 is not an ideal metric for whole metagenome shotgun data. Observed taxa is a more sensible metric. This PR replaces Chao1 with Observed taxa.

### DIFF
--- a/notebook_examples/notebooks_demo.ipynb
+++ b/notebook_examples/notebooks_demo.ipynb
@@ -54,7 +54,7 @@
    "source": [
     "## Question #1: How does alpha diversity vary by sample group?\n",
     "\n",
-    "Here, we display Chao1, Simpson’s Index, and Shannon Entropy side-by-side, grouped by the region of birth. Each group includes samples taken across the entire three-year longitudinal study."
+    "Here, we display Observed taxa, Simpson’s Index, and Shannon Entropy side-by-side, grouped by the region of birth. Each group includes samples taken across the entire three-year longitudinal study."
    ]
   },
   {
@@ -63,11 +63,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "chao1 = samples.plot_metadata(vaxis=\"chao1\", haxis=\"geo_loc_name\", return_chart=True)\n",
+    "observed_taxa = samples.plot_metadata(vaxis=\"observed_taxa\", haxis=\"geo_loc_name\", return_chart=True)\n",
     "simpson = samples.plot_metadata(vaxis=\"simpson\", haxis=\"geo_loc_name\", return_chart=True)\n",
     "shannon = samples.plot_metadata(vaxis=\"shannon\", haxis=\"geo_loc_name\", return_chart=True)\n",
     "\n",
-    "chao1 | simpson | shannon"
+    "observed_taxa | simpson | shannon"
    ]
   },
   {
--- a/notebook_examples/notebooks_demo.ipynb
+++ b/notebook_examples/notebooks_demo.ipynb
@@ -54,7 +54,7 @@
    "source": [
     "## Question #1: How does alpha diversity vary by sample group?\n",
     "\n",
-    "Here, we display Observed taxa, Simpson’s Index, and Shannon Entropy side-by-side, grouped by the region of birth. Each group includes samples taken across the entire three-year longitudinal study."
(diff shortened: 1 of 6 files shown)


onecodex/lib/enums.py
```diff
@@ -28,8 +28,9 @@ class Metric(BaseEnum):
 
 class AlphaDiversityMetric(BaseEnum):
     Simpson = "simpson"
-    Chao1 = "chao1"
+    Observed_taxa = "observed_taxa"
```
jairideout, line 31: Very minor: For enums we tend to use CamelCase instead of snake_case:

```python
ObservedTaxa = "observed_taxa"
```

onecodex/distance.py
```diff
@@ -28,9 +29,18 @@ def alpha_diversity(self, metric=AlphaDiversityMetric.Shannon, rank=Rank.Auto):
...
+            warnings.warn(
+                "`Chao1` is deprecated and will be removed in a future release. Please use `observed_taxa` instead.",
+                DeprecationWarning,
+            )
+
         df = self.to_df(rank=rank, normalize=self._guess_normalized())
 
-        output = skbio.diversity.alpha_diversity(metric, df.values, df.index, validate=False)
+        if metric == "observed_taxa":
+            output = df[df.columns].gt(0).sum(axis=1)
```
jairideout, line 41: It might be better to pass `metric="observed_otus"` to scikit-bio's `alpha_diversity` function, just to avoid reimplementing that metric here. We could still use the name "observed_taxa" in our output. The metric is super simple though, so feel free to ignore this comment.

Here's what that might look like:

```python
skbio_metric = "observed_otus" if metric == "observed_taxa" else metric
output = skbio.diversity.alpha_diversity(skbio_metric, df.values, df.index, validate=False)
```

Side note: I like the name "observed_taxa" much better than "observed_species" or "observed_otus" 🎉  -- this has always been a confusing/historical leftover from qiime1 days 😄

onecodex/distance.py
```diff
@@ -1,6 +1,7 @@
 from onecodex.exceptions import OneCodexException
 from onecodex.taxonomy import TaxonomyMixin
 from onecodex.lib.enums import AlphaDiversityMetric, BetaDiversityMetric, Rank
+import warnings
```
jairideout, line 4: Minor style comment: we tend to place imports from the standard library at the top of the file, followed by third-party imports, followed by our own `onecodex` imports (usually each of these groups are separated by a blank line). For example:

```python
import warnings

from onecodex.exceptions import OneCodexException
from onecodex.taxonomy import TaxonomyMixin
from onecodex.lib.enums import AlphaDiversityMetric, BetaDiversityMetric, Rank
```

onecodex/distance.py
clausmith: @lynchde sorry for flip-flopping - I know I told you the opposite before but I like @jairideout's solution here.

notebook_examples/notebooks_demo.ipynb
```diff
@@ -54,7 +54,7 @@
    "source": [
     "## Question #1: How does alpha diversity vary by sample group?\n",
     "\n",
-    "Here, we display Chao1, Simpson’s Index, and Shannon Entropy side-by-side, grouped by the region of birth. Each group includes samples taken across the entire three-year longitudinal study."
+    "Here, we display Observed taxa, Simpson’s Index, and Shannon Entropy side-by-side, grouped by the region of birth. Each group includes samples taken across the entire three-year longitudinal study."
```
clausmith, line 57: I'd be inclined to write "observed taxa" in lower case here, since it's not technically a proper noun? Super minor nit.

onecodex/lib/enums.py
lynchde: CamelCase it is. Thank you.

onecodex/distance.py
lynchde: Implemented @jairideout 's suggestion.

onecodex/distance.py
lynchde: Implemented.

notebook_examples/notebooks_demo.ipynb
lynchde: Good nit! Fixed.

PR #361, opened 2020-08-03
Add download retries to ResourceDownloadMixin

jairideout:
Added retries to `ResourceDownloadMixin` to support downloading files that may not be immediately available (e.g. downloading an `Assembly` FASTA file that is exported and uploaded to S3 asynchronously).

Also added support for using the server's suggested "save as" filename if the user doesn't provide a path or file object. Currently, only the `Assemblies` resource uses this new feature.

This change is necessary to support https://github.com/onecodex/mainline/pull/4057

### DIFF
--- a/onecodex/models/helpers.py
+++ b/onecodex/models/helpers.py
@@ -140,15 +140,15 @@ class ResourceDownloadMixin(object):
 
         try:
             method_to_call = getattr(self._resource, _resource_method)
-            download_info = method_to_call()
+            download_link_info = method_to_call()
 
             if path is None and file_obj is None:
                 if _filename is None:
-                    if "filename" not in download_info:
+                    if "save_as_filename" not in download_link_info:
                         raise OneCodexException(
                             "Please specify `path`, `file_obj`, or `_filename`."
                         )
-                    _filename = download_info["filename"]
+                    _filename = download_link_info["save_as_filename"]
                 path = os.path.join(os.getcwd(), _filename)
 
             if path and os.path.exists(path):
@@ -159,7 +159,7 @@ class ResourceDownloadMixin(object):
             else:
                 session = requests.Session()
 
-            link = download_info["download_uri"]
+            link = download_link_info["download_uri"]
 
             # Retry up to 5 times with backoff timing of 2s, 4s, 8s, 16s, and 32s (applies to all
             # HTTP methods). 404 is included for cases where the file is being asynchronously
--- a/onecodex/models/helpers.py
+++ b/onecodex/models/helpers.py
@@ -2,6 +2,8 @@ import click
 import inspect
 import os
 import requests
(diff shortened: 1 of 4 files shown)


onecodex/models/helpers.py
```diff
@@ -136,25 +138,47 @@ def _download(
...
+                session = self._resource._client.session
             else:
-                resp = requests.get(method_to_call()["download_uri"], stream=True)
+                session = requests.Session()
+
+            link = download_info["download_uri"]
+
+            # Retry up to 3 times with backoff timing of 4s, 8s, and 16s (applies to all HTTP
+            # methods). 404 is included for cases where the file is being asynchronously uploaded to
+            # S3 and is expected to be available soon.
```
Keats, line 166: would the 28s be enough for the 404 case?

jairideout: From testing locally it was (I put some `sleep` calls in the task to simulate delays) but it's hard to say if that'll work in the real world (e.g. if the task queue is really overloaded or something). `darwin2-ui` is using 5 retries with a backoff factor of 2. I'll update this code to match `darwin2-ui` since it's been working for ATCC downloads.

onecodex/models/helpers.py
```diff
@@ -136,25 +138,47 @@ def _download(
...
+            adapter = HTTPAdapter(max_retries=retry_strategy)
+            session.mount("http://", adapter)
+            session.mount("https://", adapter)
+
+            resp = session.get(link, stream=True)
 
             with (open(path, "wb") if path else file_obj) as f_out:
                 if progressbar:
-                    with click.progressbar(length=self.size, label=self.filename) as bar:
+                    with click.progressbar(length=self.size, label=self.id) as bar:
```
Keats, line 181: what's the reason for that change?

jairideout: `self.filename` is the original filename (e.g. `assembly.fa.gz`) and might be confusing to users since they're getting an uncompressed FASTA (e.g. `Escherichia_coli_ABC_8.fasta`). The UUID isn't super informative to the user but at least it won't be misleading regarding the filename or file type. Open to other options here -- I wanted to use the actual filename we're downloading the file to, but the user may supply a file-like object, which isn't guaranteed to have a path/filename.

Keats: Can we do:

1. see if there is the actual filename and use that
2. if not use `self.id` with a note in parenthesis explaining what is is

jairideout: I ended up using the actual filename we're downloading to if we have a path, otherwise using `self.filename` since that's probably more descriptive for users (and they are already managing their own file paths).

jairideout:
The `import-times` job is failing but it's probably unrelated (my changes don't introduce any new top-level imports). This is ready for final review!

PR #365, opened 2020-08-13
Adds aitchison distance and supports cityblock-manhattan interchangea…

lynchde:
This will allow users to calculate Aitchison distance (beta_diversity) between samples.
It also allows users to call either 'cityblock' or 'manhattan' as a beta diversity metric, as these names are synonyms.
Closes DEV-2947 and DEV-3990

### DIFF
--- a/onecodex/viz/_distance.py
+++ b/onecodex/viz/_distance.py
@@ -99,7 +99,7 @@ class VizDistanceMixin(DistanceMixin):
             Analysis will be restricted to abundances of taxa at the specified level.
         metric : {'braycurtis', 'cityblock', 'manhattan', 'jaccard', 'unifrac', 'unweighted_unifrac', 'aitchison'}, optional
             Function to use when calculating the distance between two samples.
-            Note that 'cityblock' and 'manhattan' are interchangeable metrics.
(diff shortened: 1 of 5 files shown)


onecodex/viz/_distance.py
```diff
@@ -95,8 +97,9 @@ def plot_distance(
         ----------
         rank : {'auto', 'kingdom', 'phylum', 'class', 'order', 'family', 'genus', 'species'}, optional
             Analysis will be restricted to abundances of taxa at the specified level.
-        metric : {'braycurtis', 'manhattan', 'jaccard', 'unifrac', 'unweighted_unifrac}, optional
+        metric : {'braycurtis', 'cityblock', 'manhattan', 'jaccard', 'unifrac', 'unweighted_unifrac', 'aitchison'}, optional
             Function to use when calculating the distance between two samples.
+            Note that 'cityblock' and 'manhattan' are interchangeable metrics.
```
audy, line 102: I think it's clearer to use the word equivalent instead of interchangeable

lynchde: Thank you. Changed in another commit.

PR #366, opened 2020-08-20
Improves handling of custom corporate CA bundles

clausmith:
This PR modifies the behaviour of the underlying `potion_client` to respect environment settings for prepared requests. The goal is to provide a fix to SSL verification issues encountered in certain corporate networks that are using custom CA bundles.

From the `requests` docs (https://requests.readthedocs.io/en/master/user/advanced/#prepared-requests):

> When you are using the prepared request flow, keep in mind that it does not take into account the environment. This can cause problems if you are using environment variables to change the behaviour of requests. For example: Self-signed SSL certificates specified in REQUESTS_CA_BUNDLE will not be taken into account. As a result an SSL: CERTIFICATE_VERIFY_FAILED is thrown. You can get around this behaviour by explicitly merging the environment settings into your session.

Specifically, it uses the `merge_environment_settings` function in `requests` to respect `REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE` env vars for prepared requests.

### DIFF
--- a/onecodex/vendored/potion_client/links.py
+++ b/onecodex/vendored/potion_client/links.py
@@ -93,7 +93,14 @@ class LinkBinding(object):
         req = self.request_factory(data, params)
         prepared_request = self.owner._client.session.prepare_request(req)
 
-        response = self.owner._client.session.send(prepared_request)
+        # Update cilent session so that it will properly load from env vars
+        # See https://requests.readthedocs.io/en/master/user/advanced/#prepared-requests
+        # for more details
+        settings = self.owner._client.session.merge_environment_settings(
+            url=prepared_request.url, proxies={}, stream=None, verify=None, cert=None
+        )
+
+        response = self.owner._client.session.send(prepared_request, **settings)
 
         # return error for some error conditions
         self.raise_for_status(response)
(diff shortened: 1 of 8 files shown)


onecodex/vendored/potion_client/links.py
```diff
@@ -92,6 +92,9 @@ def raise_for_status(self, response):
     def make_request(self, data, params):
         req = self.request_factory(data, params)
         prepared_request = self.owner._client.session.prepare_request(req)
+        settings = self.owner._client.session.merge_environment_settings(
```
boydgreenfield, line 95: This needs a comment as to _why_ we're modifying the session and linking to the requests docs. Also, I _think_ we should be calling this right after we instantiate the client in the main `__init__` method (off of `session`)? That's probably better even if 100% of requests flow through `make_request` in `potion_client/links` (which I'm not 100% sure of).

boydgreenfield: Ideally, we'd also have a test showing the `CA_BUNDLE...` gets loaded. We can just mock the environment in pytest.

boydgreenfield: OK I added a request to ensure that we're calling `merge_environment_settings`. It's not great, but better than missing it I suppose.

PR #371, opened 2020-10-22
Fix GitHub workflows

jairideout:
Fixes `pre-commit` and `test-python3-w-altair` GitHub workflows:

- Pins Python version at 3.8 instead of implicitly using 3.9 in `pre-commit` (core dependencies like numpy are missing wheel builds for Python 3.9)
- Pins nbconvert package below major version 6 (nbconvert 6 made some big breaking changes to [how templates work](https://nbconvert.readthedocs.io/en/latest/customizing.html))

Also removed `requirements.txt` and updated `README.md` development instructions. The file was outdated, and we want developers to use the dependencies specified in `setup.py` since that's what users and CI/CD install from.

Closes DEV-4751


boydgreenfield (approved):
👍

PR #372, opened 2020-10-22
Filter out NaNs from alpha diversity plots

jairideout:
Samples that have `NaN` as an alpha diversity value are now filtered out before plotting. This works in a similar manner to how `NaNs` are filtered when plotting numeric metadata on the y-axis.

Closes [DEV-4752](https://linear.app/onecodex/issue/DEV-4752/plot-builder-diversity-y-axis-displaying-as-factor-and-not-numeric)


jairideout:
Test failures are unrelated (see DEV-4751), this is ready for review!

clausmith (approved):

PR #381, opened 2020-12-17
Add support for concatenating multiline files

pkonieczny:
Multiline files are detected and concatenated before upload.
The multiline groups are detected based on filenames, specifically based on the `_LXXX` part, e.g. `Sample_R1_L001.fq`, `Sample_R1_L002.fq`.
If there is a gap in the sequence (e.g. [`Sample_R1_L001.fq`, `Sample_R1_L003.fq`]), the group is skipped. 
If there is a mismatch in forward and reverse file sequence (e.g. [`Sample_R1_L001.fq`, `Sample_R2_L001.fq`, `Sample_R2_L002.fq`]), the group is skipped. 
If the sequence doesn't begin with `L001`, the group is skipped.
The concatenated filename is generated by dropping the `LXXX` part of the name, e.g. [`Sample_R1_L001.fq`, `Sample_R1_L002.fq`] &rarr; `Sample_R1.fq`.
The concatenated files are put into a temporary directory.
The concatenated files replace the multiline files on the input list.

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -306,7 +306,8 @@ class AnalysisMixin(
 
         if normalize is True or (normalize == "auto" and rank):
             if not self._guess_normalized():
-                df = df.div(df.sum(axis=1), axis=0)
+                # Replace nans with zeros for samples that have a total abundance of zero.
+                df = df.div(df.sum(axis=1), axis=0).fillna(0.0)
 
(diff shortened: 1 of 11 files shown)


boydgreenfield:
@clausmith Can you review?

clausmith (changes requested):
One minor change to the prompt wording but otherwise looks good to me!

PR #408, opened 2021-11-24
Test on python 3.9

Keats:
## Status
- [x] **Ready**

## Description
Also drop python 3.6 as it's EOL at the end of the year

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,9 +13,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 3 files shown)


jairideout (approved):

jairideout:
Looks like the test matrix is still expecting 3.6 results for macOS & Linux?

Keats:
I believe it's a bug in GH actions, we ran into the same one for chrononaut

jairideout:
Weird... we can't merge the PR because those two statuses are required. Maybe try closing and reopening the PR?

Keats:
I can as an admin. @pkonieczny even had issues on a subsequent PR, it's really odd.

PR #414, opened 2022-01-07
Python 3.10 support; scikit-bio upgrade; additional cleanup and docs

audy:
## Status

- [x] **Ready**

## Description

- Added Python 3.10 support
- Updated scikit-bio to >=0.5.7, which no longer requires numpy to be installed first and supports Python 3.10 and Apple M1
- Removed `tox`: it had drifted from the Makefile and GitHub Actions, and we no longer install testing dependencies as part of the package in favor of `pre-commit`
- Additional cleanup related to Python 2 removal
- Updated setup.py and readme to reflect supported Python versions

Closes DEV-6474

## Related PRs

- [x] This PR is independent


jairideout:
@audy I reran the tests and they're failing because scikit-bio doesn't support Python 3.10 yet. There's an open issue (https://github.com/biocore/scikit-bio/issues/1791) and PR (https://github.com/biocore/scikit-bio/pull/1798) that last had activity 20 days ago. We could wait for a scikit-bio release that supports Python 3.10, or just add `test-python3.10-minimal` to the build matrix (e.g. to test the non-skbio parts of this package).

Keats:
Yep @jairideout that's the reason why mainline is using 3.9 rather than 3.10 as well.

Keats:
It looks like the latest scikit-bio does support 3.10 so we can re-run the tests after rebasing once https://github.com/onecodex/onecodex/pull/420 is merged

Keats (approved):

PR #415, opened 2022-03-14
Add support for downloading batches of samples

jairideout:
## Status
- [x] **Ready**

## Description
Added `download_samples()` function and `onecodex download samples` CLI command for downloading batches of samples as FASTA/Q files:
- All samples in your account
- Samples optionally filtered by project name or ID
- Samples optionally filtered by tag names

The code was ported from: https://gist.github.com/clausmith/c5da3145412265014151cfcad5493180

Closes DEV-6563

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/lib/download.py
+++ b/onecodex/lib/download.py
@@ -1,0 +1,75 @@
+import logging
+import os
+import os.path
+import warnings
+
+import click
+
+from onecodex.utils import FakeProgressBar
+from onecodex.exceptions import OneCodexException
+
+log = logging.getLogger("onecodex")
+
+
+def get_samples_from_project(ocx, project_name_or_id):
+    project = ocx.Projects.get(project_name_or_id)
+
+    if not project:
+        try:
+            project = ocx.Projects.where(name=project_name_or_id)[0]
+        except IndexError:
+            raise OneCodexException(
+                "No project exists with name or ID '{}'.".format(project_name_or_id)
+            )
+
+    return ocx.Samples.where(project=project)
+
+
+def filter_samples_by_tags(ocx, samples, tag_names):
+    tags = set()
+    for tag_name in tag_names:
+        log.info("Filtering samples by tag '{}'...".format(tag_name))
+        try:
+            tag = ocx.Tags.where(name=tag_name)[0]
+        except IndexError:
+            warnings.warn("No tag found: '{}'.".format(tag_name))
+            continue
+        tags.add(tag)
+
+    if tags:
+        # Filter for samples with any of the tags above (union)
+        return [sample for sample in samples if len(set(sample.tags) & tags)]
+    return samples
+
+
+def download_samples(ocx, outdir, project_name_or_id=None, tag_names=None, progressbar=False):
+    if project_name_or_id:
+        log.info("Fetching samples in project '{}'...".format(project_name_or_id))
+        samples = get_samples_from_project(ocx, project_name_or_id)
+    else:
+        log.info("Fetching all samples in your account...")
+        samples = ocx.Samples.all()
+
+    if tag_names:
+        samples = filter_samples_by_tags(ocx, samples, tag_names)
+
+    if not len(samples):
+        log.info("No samples match the filter criteria; nothing to download.")
+        return []
+
+    if not os.path.exists(outdir):
+        os.makedirs(outdir)
+
+    if progressbar:
+        label = "Downloading {} sample{}...".format(len(samples), "" if len(samples) == 1 else "s")
+        progressbar = click.progressbar(length=len(samples), label=label)
+    else:
+        progressbar = FakeProgressBar()
+
+    filepaths = []
+    with progressbar as bar:
+        for sample in samples:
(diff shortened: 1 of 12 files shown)


tests/test_speed.py
```diff
@@ -25,7 +25,7 @@
             "from onecodex.viz import VizPCAMixin",
             {
                 "onecodex": 0.25,
-                "onecodex.viz": 0.20,
+                "onecodex.viz": 0.25,
```
jairideout, line 28: This is unrelated to the changes in this PR. I've noticed this test fails quite frequently, let's see if bumping the timeout helps.

onecodex/lib/download.py
```diff
@@ -0,0 +1,75 @@
...
+        label = "Downloading {} sample{}...".format(len(samples), "" if len(samples) == 1 else "s")
+        progressbar = click.progressbar(length=len(samples), label=label)
+    else:
+        progressbar = FakeProgressBar()
+
+    filepaths = []
+    with progressbar as bar:
+        for sample in samples:
+            filepath = os.path.join(outdir, sample.filename)
+            filepaths.append(sample.download(path=filepath, progressbar=False))
```
clausmith, line 73: Users can only download samples with `tier == "premium"`. We should catch the exception, log a message noting that the sample could not be downloaded and to contact us regarding your subscription, and skip the sample.

clausmith (changes requested):
We should gracefully handle the case where users try and download non-premium tier samples.

onecodex/lib/download.py
jairideout: @clausmith here's how the warning message looks like from the CLI now:

> WARNING: Skipping download of sample 92cc98f50fdc4855 to output/test-data.fasta.gz: You must either have a premium platform account or be in a notebook environment to download files. Please feel free to contact us  about your subscription at support@onecodex.com.

onecodex/lib/download.py
```diff
@@ -0,0 +1,80 @@
...
+    if progressbar:
+        label = "Downloading {} sample{}...".format(len(samples), "" if len(samples) == 1 else "s")
+        progressbar = click.progressbar(length=len(samples), label=label)
+    else:
+        progressbar = FakeProgressBar()
+
+    filepaths = []
+    with progressbar as bar:
+        for sample in samples:
+            filepath = os.path.join(outdir, sample.filename)
```
jairideout, line 72: @clausmith When a file already exists in the output directory, we skip the download and warn the user. Should we include the sample UUID somewhere in the downloaded filename? e.g. `{sample.filename}_{uuid}.fastq.gz` or `{uuid}_{sample.filename}.fastq.gz`

That could help users know exactly which sample was downloaded in case multiple samples have the same filename.

onecodex/lib/download.py
```diff
@@ -0,0 +1,80 @@
...
+            try:
+                filepaths.append(sample.download(path=filepath, progressbar=False))
+            except OneCodexException as e:
+                warnings.warn(
+                    "Skipping download of sample {} to {}: {}".format(sample.id, filepath, str(e))
+                )
+            bar.update(1)
+    return filepaths
```
Keats, line 80: is it easy to mock the ocx calls? Those functions seem like good unit tests targets.

onecodex/lib/download.py
```diff
@@ -0,0 +1,80 @@
...
+    return samples
+
+
+def download_samples(ocx, outdir, project_name_or_id=None, tag_names=None, progressbar=False):
+    if project_name_or_id:
+        log.info("Fetching samples in project '{}'...".format(project_name_or_id))
+        samples = get_samples_from_project(ocx, project_name_or_id)
+    else:
+        log.info("Fetching all samples in your account...")
+        samples = ocx.Samples.all()
```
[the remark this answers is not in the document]
jairideout: Yeah it executes the API query. We let the user know what's happening in the logging message above, but you're right this might be a problem for big accounts.

@clausmith this is how the original download script worked (downloading all samples in the account if no filters provided). Should we enforce that the user provides at least one project or tag to filter by?

onecodex/lib/download.py
jairideout: mocked and added more tests

onecodex/lib/download.py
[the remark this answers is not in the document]
clausmith: No, there's a fairly common user story where a new user (e.g. a new Transnetyx customer) just wants to download all of the data from their first pilot project/sequencing run. If `len(samples) > n` though (for something like `n=50`), we should probably pop a confirmation prompt that says "Are you sure you'd like to download all `n` samples in your One Codex account?"

onecodex/lib/download.py
clausmith: The former (`{sample.filename}_{uuid}.fastq.gz`) sounds reasonable to me.

onecodex/lib/download.py
clausmith: Oh perfect - looks great.

clausmith (approved):

PR #417, opened 2022-03-17
Fix `metric` parameter behavior in SampleCollection constructor

jairideout:
## Status
- [x] **Ready**

## Description
Previously when constructing a `SampleCollection`, the only way to override the default metric of "auto" was via the `field` parameter, which is deprecated. Anything passed via `metric` was ignored, *unless* `field` was set to something false-y.

Now the constructor uses `metric` correctly, and disallows passing both `metric` and `field`. `field` is aliased to `metric` and raises a `DeprecationWarning`.

Closes DEV-6052

## Related PRs
- [x] This PR is independent


alenajk (approved):
👍

Keats (approved):

boydgreenfield:
🥳

PR #420, opened 2022-03-30
Update deps and drop python2

Keats:
## Status
- [ ] **Ready**
- [x] **In progress**

## Description
Drop python2 support
Upgrade all dependencies and see what breaks

### DIFF
--- a/onecodex/notebooks/exporters.py
+++ b/onecodex/notebooks/exporters.py
@@ -50,7 +50,6 @@ class AltairPreprocessor(Preprocessor):
 
 class OneCodexHTMLExporter(HTMLExporter):
     export_from_notebook = "One Codex HTML Report"
-    template_path = [ASSETS_PATH]
 
     def __init__(self, config=None, **kw):
         super(OneCodexHTMLExporter, self).__init__(config=config, **kw)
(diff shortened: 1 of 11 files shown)


jwin4740:
New to this project. Is there someone I can contact for help onboarding?

onecodex/notebooks/exporters.py
```diff
@@ -70,6 +69,11 @@ def __init__(self, config=None, **kw):
         except ImportError:
             return
 
+    @property
+    def template_paths(self):
+        # https://github.com/jupyter/nbconvert/issues/1492
```
bcamarda, line 74: should we add this comment before line 172, too?

setup.py
```diff
@@ -38,18 +38,25 @@
     "vega_datasets",
     "black",
```
jairideout, line 39: Can you pin `black` to 22.3.0 so that it matches pre-commit?

jairideout (approved):

setup.py
jairideout: Great idea! 👍

jairideout (approved):

PR #424, opened 2022-06-02
to_df method update for functional runs

safay:
## Status
- [x] **Ready**

## Description
Add functionality to SampleCollection objects, to generate tabular data for Functional runs using the `.to_df()` method of a `SampleCollection`. This aims to be a pattern to be replicated for any analysis job type that a user might want to compare results for across samples, including control runs, alignments, etc..

## Related PRs
- [x] This PR depends on changes to the Mainline API:
  -  https://github.com/onecodex/mainline/pull/5290


jairideout:
I'm going to try closing and reopening the PR to see if it helps the diff situation

jairideout:
No luck :\

jairideout:
@safay I managed to fix the diff by doing a `git pull --rebase origin master` locally, fixing a bunch of rebase commit conflicts, then force-pushing the rebased branch. I'm still not sure *how* the diff/commits got this way, but the diff looks correct now!

onecodex/models/collection.py
```diff
@@ -389,6 +391,171 @@ def taxonomy(self):
...
+        elif metric not in ["cpm", "rpk"]:
+            raise ValueError(
+                f"if using annotation={annotation}, 'value' must be one of ['cpm', 'rpk']"
+            )
+
+        data = {}
+        all_features = set()
+        # iterate over functional profiles, subset data, and store in data dict
+        for profile in self._functional_profiles:
+            # get table from mainline
```
jairideout, line 508: nit: since this is a public repo, we try not to refer (too much) to internal component names like mainline. it'd be better to say something like "get table using One Codex API"

onecodex/models/collection.py
```diff
@@ -389,6 +391,171 @@ def taxonomy(self):
...
+
+        # initialize an array and fill it
+        array = np.full(shape=(len(data), len(features_to_ix)), fill_value=np.nan)
+        profile_ids = []
+        for profile_index, profile_id in enumerate(data):
+            for feature_id in data[profile_id]:
+                array[profile_index, features_to_ix[feature_id]] = data[profile_id][feature_id]
+            profile_ids.append(profile_id)
+        df = pd.DataFrame(
+            array, index=pd.Index(profile_ids, name="functional_profile_uuid"), columns=feature_list
```
jairideout, line 536: usually in the client library we refer to UUIDs as "IDs", so `"functional_profile_id"` would be more consistent with other parts of the library

onecodex/models/collection.py
```diff
@@ -389,6 +391,171 @@ def taxonomy(self):
...
+            "filler": filler,
+        }
+
+    def _functional_results(self, **kwargs):
+        if "functional_results" not in self._cached:
+            self._collate_functional_results(**kwargs)
+        # check for diff in kwargs from previous call
+        for k in kwargs:
+            if self._cached["functional_results_content"].get(k) != kwargs[k]:
+                self._collate_functional_results(**kwargs)
```
jairideout, line 556: this will call `_collate_functional_results()` once for each kwarg that differs, when it really only needs to be called once. also, if `**kwargs` is a *subset* of the cached kwargs, it may not trigger a cache refresh when it should have

onecodex/models/experimental.py
```diff
@@ -158,14 +158,36 @@ def results(self, json=True):
...
         -------
-        table : `pd.DataFrame`
+        results_df : `pd.DataFrame`
             A Pandas DataFrame of the functional results.
         """
         import pandas as pd
 
-        return pd.DataFrame(self._results()["table"])
+        results_df = pd.DataFrame(self._results()["table"])
+        if annotation != "all":
```
jairideout, line 180: we should probably validate `annotation` is one of the supported choices

onecodex/viz/__init__.py
```diff
@@ -120,7 +120,7 @@ def stderr_filter(line):
 
         altair_module.renderers.enable(
             "altair_saver",
-            fmts=["html", "svg"],
```
jairideout, line 123: can you double-check with @clausmith whether the SVG saver is needed for anything? Not sure if removing it will break anything downstream

jairideout (changes requested):

onecodex/models/collection.py
```diff
@@ -389,6 +391,171 @@ def taxonomy(self):
...
+        features_to_ix = {}
+        feature_list = []
+        for ix, feature in enumerate(all_features):
+            features_to_ix[feature] = ix
+            feature_list.append(feature)
+
+        # initialize an array and fill it
+        array = np.full(shape=(len(data), len(features_to_ix)), fill_value=np.nan)
+        profile_ids = []
+        for profile_index, profile_id in enumerate(data):
```
jairideout, line 531: this is a theoretical concern: you're using the item insert order for `data` dict to determine the order of functional profiles in the result. In Python >=3.5, this will work because `dict` maintains insertion order. But it's usually a good idea to use `OrderedDict` if you are relying on dictionary order. In this case, you could use `self._functional_profiles` list to determine the row order in the dataframe

see https://gandenberger.org/2018/03/10/ordered-dicts-vs-ordereddict/

onecodex/analyses.py
```diff
@@ -239,7 +239,56 @@ def _metadata_fetch(self, metadata_fields, label=None):
...
+        kwargs : 'dict', optional
+             Keyword arguments specific to the analysis_type
+        """
+        generate_df = {
+            "classification": self.to_classification_df,
+            "functional": self.to_functional_df,
+        }
+        return generate_df[analysis_type](**kwargs)
+
+    def to_functional_df(
```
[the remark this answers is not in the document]
safay: done

onecodex/analyses.py
```diff
@@ -239,7 +239,56 @@ def _metadata_fetch(self, metadata_fields, label=None):
...
+        analysis_type : {'classification', 'functional'}, optional
+            The analysis_type to aggregate, corresponding to AnalysisJob.analysis_type
+        kwargs : 'dict', optional
+             Keyword arguments specific to the analysis_type
+        """
+        generate_df = {
+            "classification": self.to_classification_df,
+            "functional": self.to_functional_df,
+        }
+        return generate_df[analysis_type](**kwargs)
```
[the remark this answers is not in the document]
safay: By addressing your comment about adding an enum for AnalysisType, calling the method with an invalid value will raise an error like:
`ValueError: 'foo' is not a valid AnalysisType`

onecodex/models/collection.py
safay: done

onecodex/models/collection.py
safay: Oh, good catch, this is sloppy.

Fixed, added tests.
@jairideout I'm curious to hear your feedback on the way the defaults for the args are defined both in the caller and the function definition. This feels wrong to me. Maybe I should take the defaults out of the function definition and just "patch" the kwargs passed by the caller `_functional_results()`? Would appreciate your advice on a better pattern here.

onecodex/models/experimental.py
safay: done, using the enum

onecodex/viz/__init__.py
safay: This change is related to `test_reports.py` beginning to fail in the middle of working on this branch. I suspect it's related to a dependency that changed. I spent a long time banging my head against the wall trying to fix it. This is what eventually worked, and I can't really explain why.
You're right that we need to check in with @clausmith about its use downstream. One question I have is how the `onecodex` library used in report generation. This is an integration testing issue; we should not assume we can rely on `test_reports.py` to cover all cases if the associated library code is used for production downstream.

onecodex/models/collection.py
safay: Is your concern specifically that the order of the rows in the returned dataframe may not match the order of the functional profiles in `self._functional_profiles()` (if we're using python <=3.5)? Or is it that there is possibly a data integrity issue? The index of the df is set by `profile_ids`, generated inside the enumeration of `data`, so if I'm reading it correctly, the data table should be intact, though rows could be in a different order than the profiles in `._functional_profiles()`.

onecodex/models/collection.py
safay: fixed

onecodex/analyses.py
```diff
@@ -239,7 +239,56 @@ def _metadata_fetch(self, metadata_fields, label=None):
...
+        metric="coverage",
+        fill_missing=False,
+        filler=0,
+    ):
+        """
+        Return a dataframe of results from a functional analysis.
+
+        Parameters
+        ----------
+        annotation : {'pathways', 'metacyc', 'eggnog', 'go', 'ko', 'ec', 'pfam', 'reaction'}, optional
```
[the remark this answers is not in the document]
safay: done

onecodex/analyses.py
```diff
@@ -239,7 +239,56 @@ def _metadata_fetch(self, metadata_fields, label=None):
...
-    def to_df(
+    def to_df(self, analysis_type="classification", **kwargs):
+        """
+        Transform Analyses of samples in a SampleCollection into tabular format.
+
+        Parameters
+        ----------
+        analysis_type : {'classification', 'functional'}, optional
+            The analysis_type to aggregate, corresponding to AnalysisJob.analysis_type
+        kwargs : 'dict', optional
```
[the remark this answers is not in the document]
safay: fixed

onecodex/analyses.py
```diff
@@ -239,7 +239,56 @@ def _metadata_fetch(self, metadata_fields, label=None):
...
+             Keyword arguments specific to the analysis_type
```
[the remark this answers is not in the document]
safay: I'm worried that this will become unwieldy as we add more `analysis_type`s. Added advice to refer to individual function definitions, do you think that is sufficient?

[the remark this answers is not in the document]
jairideout: 👍

onecodex/viz/__init__.py
jairideout: a little more info on this: https://github.com/onecodex/onecodex/pull/426#discussion_r911546693

onecodex/models/collection.py
jairideout: yeah I was concerned the order of rows in the df *could* differ from the order of `self._functional_profiles`. Not sure if that'll cause any confusion/issues downstream, I just remember that the ordering of classifications matches the df's row order in `_collate_results()` and thought it might be safer to do that here too. Feel free to ignore because it's probably just a theoretical issue or a non-issue

onecodex/lib/enums.py
```diff
@@ -89,3 +89,19 @@ class Linkage(BaseEnum):
...
+
+class AnalysisType(BaseEnum):
+    Classification = "classification"
+    Functional = "functional"
+
+
+class FunctionalAnnotations(BaseEnum):
+    Pathways = "pathways"
+    MetaCyc = "metacyc"
+    EggNOG = "eggnog"
```
jairideout, line 102: style nit: for the "keys" in enums, we try to use camel-case that handles acronyms like this:

```
EggNog
Go
Ko
Ec
```

I know this looks weird but it'd match the naming styling of other things in the library and in mainline (e.g. OrdinationMethod.Pcoa, OrdinationMethod.Smacof)

onecodex/analyses.py
```diff
@@ -239,7 +239,56 @@ def _metadata_fetch(self, metadata_fields, label=None):
...
+            AnalysisType.Classification: self._to_classification_df,
+            AnalysisType.Functional: self._to_functional_df,
+        }
+        return generate_df[AnalysisType(analysis_type)](**kwargs)
+
+    def _to_functional_df(
+        self,
+        annotation=FunctionalAnnotations.Pathways,
+        taxa_stratified=True,
+        metric="coverage",
```
jairideout, line 263: can you make `metric` an enum too? I'm liking the free validation we get with the other enums you added 👍

onecodex/models/collection.py
```diff
@@ -389,18 +392,181 @@ def taxonomy(self):
...
+        self._cached["functional_results"] = df
+        self._cached["functional_results_content"] = {
+            "annotation": annotation,
+            "taxa_stratified": taxa_stratified,
+            "metric": metric,
+            "fill_missing": fill_missing,
+            "filler": filler,
+        }
+
+    def _functional_results(self, **kwargs):
```
jairideout, line 542: Here's how I'd suggest handling the kwarg matching:

1. use this method signature (i.e. make all parameters *required positional arguments*):
    ```python
    def _collate_functional_results(self, annotation, metric, taxa_stratified, fill_missing, filler):
        ...
    ```
2. `_functional_results()` becomes:
    ```python
    def _functional_results(self, **kwargs):
        if "functional_results" not in self._cached:
            self._collate_functional_results(**kwargs)
            return self._cached["functional_results"]

        if kwargs != self._cached["functional_results_content"]:
            self._collate_functional_results(**kwargs)

        return self._cached["functional_results"]
    ```

The benefits of this approach are:
 - we only define the default values once (in `_to_functional_df()`)
 - if a new parameter gets added, it only has to be added to `_to_functional_df()` signature and to `_collate_functional_results()`
 - if we *forget* to pass a parameter through to `_collate_functional_results()`, it'll error because all arguments are required

jairideout (changes requested):

onecodex/models/collection.py
jairideout: see https://github.com/onecodex/onecodex/pull/424#discussion_r919136680 for an idea I had

onecodex/viz/__init__.py
safay: I added a comment to the thread on Gerrit's PR

onecodex/models/collection.py
safay: This significantly simplifies things. I definitely got wrapped up in my own historical spaghetti as I developed these methods, thanks for helping me see past it!

onecodex/analyses.py
safay: done

jairideout (approved):
Thanks for the updates @safay! I'm approving the PR but let's wait on merging until we figure out the failing `test_reports.py` and whether removing the SVG altair renderer is an ok solution: https://github.com/onecodex/onecodex/pull/424#discussion_r915153146

audy:
I was able to get `tests/test_reports.py::test_html_report_generation` to work with svg by using the nodejs back-end for altair_saver, following these instructions: tests/test_reports.py::test_html_report_generation

onecodex/viz/__init__.py
jairideout: fixed upstream and rebased

PR #426, opened 2022-06-29
Set median box plot line stroke to black

GSGerritsen:
## Status
- [x] **Ready**

## Description
One possible solution to [DEV-7001](https://linear.app/onecodex/issue/DEV-7001/custom-plots-box-plot-bug-alpha-diversity-group-by) - if we set the stroke color for the median line in the box plot, then at least a visible horizontal line will display for this edge case.

How the edge case mentioned in the linear issue would look:
Screen Shot 2022-06-28 at 10 13 37 PM

How it would affect normal boxes:
Screen Shot 2022-06-28 at 10 12 53 PM

**Alternate solution**: if we don't want to change the aesthetics of the box plot (current can be seen in linear issue), we could issue a warning at the top of the page along the lines of "box won't render on duplicate data". We do a similar warning for an empty box already.

Also fixed a random typo I found ("categroy" -> "category")

cc @clausmith @jairideout for thoughts

### DIFF
--- a/onecodex/viz/__init__.py
+++ b/onecodex/viz/__init__.py
@@ -120,7 +120,7 @@ def configure_onecodex_theme(altair_module=None):
 
         altair_module.renderers.enable(
             "altair_saver",
-            fmts=["html", "svg"],
+            fmts=["html"],
             embed_options=VEGAEMBED_OPTIONS,
             vega_cli_options=["--loglevel", "error"],
             stderr_filter=stderr_filter,
--- a/onecodex/viz/__init__.py
+++ b/onecodex/viz/__init__.py
@@ -68,7 +68,7 @@ def onecodex_theme():
         "config": {
             "range": {
                 "heatmap": list(reversed(onecodex_palette)),
-                "categroy": DEFAULT_PALETTES["ocx"],
+                "category": DEFAULT_PALETTES["ocx"],
                 "ramp": list(reversed(onecodex_palette)),
             },
             "area": {"fill": OCX_DARK_GREEN},
--- a/onecodex/viz/__init__.py
+++ b/onecodex/viz/__init__.py
@@ -120,7 +120,7 @@ def configure_onecodex_theme(altair_module=None):
 
         altair_module.renderers.enable(
             "altair_saver",
-            fmts=["html"],
+            fmts=["html", "svg"],
             embed_options=VEGAEMBED_OPTIONS,
             vega_cli_options=["--loglevel", "error"],
             stderr_filter=stderr_filter,
--- a/tests/test_viz.py
+++ b/tests/test_viz.py
@@ -6,7 +6,7 @@ pytest.importorskip("numpy")  # noqa
 pytest.importorskip("pandas")  # noqa
 
 import numpy as np
-
+import math
 from onecodex.exceptions import OneCodexException, PlottingException, PlottingWarning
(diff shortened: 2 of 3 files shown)


tests/test_viz.py
```diff
@@ -93,7 +93,7 @@ def test_plot_metadata_alpha_diversity_with_nans(ocx, api_data):
     with mock.patch.object(SampleCollection, "alpha_diversity", return_value=mock_alpha_div_values):
         chart = samples.plot_metadata(vaxis="shannon", return_chart=True)
 
-    assert chart.data["shannon"].tolist() == [5.2120437645419395]
+    assert chart.data["shannon"].tolist() == [5.212043764541939]
```
petraszd, line 96: That is example of a bad test. Comparing floats is dangerous. Probably CPU acts slightly differently on ARM versus x86.

```python
shannon_result = chart.data["shannon"].tolist()
assert len(shannon_result) == 1
assert math.isclose(shannon_result[0], 5.212043764541939)
# Or
assert shannon_result[0] == pytest.approx(5.212043764541939)
```

GSGerritsen:
> @GSGerritsen nice fix! I'm personally okay with the new aesthetics of the boxplots. Would you mind testing what it looks like when only one sample (i.e. data point) is present in the boxplot? I wonder if your fix will make the median line display for that case too. It's not necessary in order to merge this, but it'd be nice to check because we could remove that "empty box" warning and some data munging code.

@jairideout I think that is what will happen:
Screen Shot 2022-06-30 at 9 54 48 AM

I could make an issue to remove the code you mentioned? I see the one block in `plot_metadata` but not sure what else would need to be removed so probably easier to do in a separate, small PR.

tests/test_viz.py
GSGerritsen: Thanks for catching this! Updated to your suggestion.

jairideout:
@GSGerritsen 

> I could make an issue to remove the code you mentioned? I see the one block in plot_metadata but not sure what else would need to be removed so probably easier to do in a separate, small PR.

Thanks for testing it out! A separate issue/PR is fine. It should just be the code in `plot_metadata()`, and maybe a unit test that needs to be updated since the warning won't be raised anymore.

onecodex/viz/__init__.py
```diff
@@ -120,7 +120,7 @@ def stderr_filter(line):
 
         altair_module.renderers.enable(
             "altair_saver",
-            fmts=["html", "svg"],
```
GSGerritsen, line 123: @clausmith wanted to check with you on this one: a test was failing on this PR and Jai pointed this out to me as something Scott is also doing to get CI on his PR to pass. Do you know if it's okay to remove this here?

safay: I encountered a similar failing test, and spent a bit of time down a rabbit hole with googling and trial and error wrangling altair and notebook rendering deps, but couldn't resolve it with updating deps. In the end, the only thing I could get to work was changing this line to `fmts=["html"]`, which allowed the tests to pass.

The test that was failing for me was `test_html_report_generation()` in `tests/test_reports.py`, looking for the `vega-embed` text in the html doc.

@clausmith How can we update the PDF render functionality tests to make sure they cover the cases you're thinking of?

jairideout: fixed upstream and rebased

jairideout (approved):

PR #434, opened 2022-10-25
return additional data for distance plot

alenajk:
## Status
- [x] **Ready**

## Description
For custom plots exports, we need additional data to be passed from the client library to mainline for use in generating csv/xlsx formatted plot data

## Related PRs
- [x] Please see the following related PRs:
Mainline PR to follow

### DIFF
--- a/onecodex/viz/_distance.py
+++ b/onecodex/viz/_distance.py
@@ -82,7 +82,7 @@ class VizDistanceMixin(DistanceMixin):
             "labels_in_order": labels_in_order,
         }
 
-    def plot_distance(
+    def _plot_distance(
         self,
         rank=Rank.Auto,
         metric=BetaDiversityMetric.BrayCurtis,
@@ -96,7 +96,8 @@ class VizDistanceMixin(DistanceMixin):
         width=None,
         height=None,
     ):
-        """Plot beta diversity distance matrix as a heatmap and dendrogram.
+        """'Private' method that accompanies `plot_distance` so we can return chart data as well as
+        concat_chart, for use in creating data exports.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


onecodex/viz/_distance.py
```diff
@@ -221,7 +221,7 @@ def plot_distance(
         open_links_in_new_tab(concat_chart)
 
         if return_chart:
-            return concat_chart
+            return concat_chart, chart.data
```
jairideout, line 224: This'll break backwards compatibility for existing users of the client library. Is there a way to access the data with `concat_chart.data`?

alenajk: @jairideout Hmm..Unfortunately no, it's not available via `concat_chart`. I suppose we could add a method that is a duplicate of this method that returns `chart.data` instead, but that's a lot of duplicate code. Do you have other ideas for how best to handle it?

jairideout: I would double-check with @clausmith or @Keats on this, but we probably don't want to make a backwards incompatible change to the `plot_distance()` method because things could stop working for existing users (e.g. in notebooks or their own custom workflows).

One way to solve this is by moving all of the logic in `plot_distance()` into a "private" method (e.g. `_plot_distance()`) that returns `concat_chart` and `chart`. Then `plot_distance()` can call `_plot_distance()` and the logic won't be duplicated. On the Custom Plots side, we could use `_plot_distance()` to get the data we need for exports.

alenajk: @jairideout That sounds good to me.
@Keats @clausmith How does the above sound to you both?

alenajk: Update: I found a way to access the data, actually! So this change isn't needed anymore.

alenajk:
Found a way to access the data without making any client library changes after all

PR #438, opened 2022-12-13
Add test for visual differences in report generation

audy:
## Status

- [x] **Ready**

## Description

* Add Github workflow to run a visual diff on generated PDF report. The workflow will now fail if the generated report PDF has visual changes. A PDF for the expected, generated and diff is now uploaded as a GitHub artifact during test runs
* Add `ONE_CODEX_DO_NOT_INSERT_DATE` option to report generation (needed to skip inserting the date which would break visual diffs)

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Tests


notebook_examples/example.ipynb
```diff
@@ -35,14 +37,19 @@
...
-    "    color=\"lightblue\",\n",
-    "    interpolate='step-after',\n",
-    "    line=True\n",
-    ").encode(\n",
-    "    x='date',\n",
-    "    y='price'\n",
-    ").transform_filter(alt.datum.symbol == 'GOOG')\n"
+    "\n",
+    "with warnings.catch_warnings():\n",
+    "    warnings.filterwarnings(\"ignore\", message=\"iteritems is deprecated and will be removed in a future version. Use .items instead.\")\n",
```
audy, line 42: had to catch this because it shows up in the report. I could add the warning to the ground truth but the warning has a timestamp in it that always shows up in the diff.

onecodex/notebooks/exporters.py
```diff
@@ -91,8 +91,9 @@ def from_notebook_node(self, nb, resources=None, **kw):
         elif "metadata" not in resources:
             resources["metadata"] = {}
 
+        do_not_insert_date = os.environ.get("ONE_CODEX_DO_NOT_INSERT_DATE") == "True"
```
alenajk, line 94: Double negative could be confusing - would it be clearer to just use INSERT_DATE instead of DO_NOT_INSERT_DATE?

audy: Good point. Fixed.

tests/test_reports.py
```diff
@@ -51,3 +87,8 @@ def test_pdf_report_generation(capsys, nb, nb_config):
 
     # Check no stderr from vega CLI or similar
     assert capsys.readouterr().err == ""
+
+    # copy test report to $pwd so that it can be uploaded to github as an
+    # artifact
+    with open("test-report-generated.pdf", "wb") as handle:
```
jairideout, line 93: can you add `test-report-generated.pdf` to `.gitignore`? might want to also add `test-report-diff.pdf` in case someone runs the differ locally

jairideout (approved):

PR #441, opened 2023-04-04
add retry/backoff for 429s

alenajk:
## Status
- [x] **Ready**

## Description
Add retry and backoff for 429s

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/api.py
+++ b/onecodex/api.py
@@ -43,7 +43,6 @@ class Api(object):
         load_extensions=True,
         **kwargs
     ):
-
         if base_url is None:
             base_url = os.environ.get("ONE_CODEX_API_BASE", "https://app.onecodex.com")
             if base_url != "https://app.onecodex.com":
@@ -111,8 +110,8 @@ class Api(object):
             total=3, backoff_factor=4, method_whitelist=False, status_forcelist=[429]
         )
         adapter = HTTPAdapter(max_retries=retry_strategy)
-        self._session.mount('http://', adapter)
-        self._session.mount('https://', adapter)
+        self._session.mount("http://", adapter)
+        self._session.mount("https://", adapter)
 
         self._copy_resources()
 
--- a/onecodex/api.py
+++ b/onecodex/api.py
@@ -107,7 +107,7 @@ class Api(object):
 
         # Set backoff / retry strategy for 429s
         retry_strategy = Retry(
-            total=3, backoff_factor=4, method_whitelist=False, status_forcelist=[429]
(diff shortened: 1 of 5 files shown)


onecodex/api.py
```diff
@@ -103,6 +105,15 @@ def __init__(
         )
         self._client._fetch_schema(cache_schema=cache_schema)
         self._session = self._client.session
+
+        # Set backoff / retry strategy for 429s
+        retry_strategy = Retry(
+            total=3, backoff_factor=4, method_whitelist=False, status_forcelist=[429]
```
jairideout, line 111: We should probably add 502 and 503 codes to the list

onecodex/api.py
```diff
@@ -103,6 +105,15 @@ def __init__(
         )
         self._client._fetch_schema(cache_schema=cache_schema)
         self._session = self._client.session
+
+        # Set backoff / retry strategy for 429s
+        retry_strategy = Retry(
```
jairideout, line 110: I think the retry code in `onecodex/lib/upload.py` can be removed now that we have retries being set here

onecodex/api.py
```diff
@@ -103,6 +105,15 @@ def __init__(
         )
         self._client._fetch_schema(cache_schema=cache_schema)
         self._session = self._client.session
+
+        # Set backoff / retry strategy for 429s
```
jairideout, line 109: It'd be nice to have a unit test for the retry logic. `test_callback_retry_handling()` could be used as an example (it's testing upload retries)

jairideout: Actually if we remove the retry logic from `onecodex/lib/upload.py`, the existing test is probably good enough

alenajk: So, the test was using a mocked version of `session` so it was failing when I removed retry code in `onecodex/lib/upload.py` - I replaced that test with a test in `test_api_models` though

tests/test_api_models.py
```diff
@@ -23,7 +23,17 @@ def test_api_creation(api_data):
...
+
+
+def test_retries_set_on_client_session(api_data):
+    ocx = Api(
+        api_key="1eab4217d30d42849dbde0cd1bb94e39",
+        base_url="http://localhost:3000",
+        cache_schema=False,
+    )
+
+    assert ocx._session.adapters["http://"].max_retries.total == 3
```
jairideout, line 35: nit: it'd be nice to also test that it's set for https (that's the one we care the most about) and that `ocx._client.session` has retries set (that object path is used by the upload module & other places in the library)

jairideout (approved):

PR #442, opened 2023-04-04
fix pandas version issues

alenajk:
## Status
- [x] **Ready**

## Description
Fixing pandas version errors

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -367,7 +367,7 @@ class AnalysisMixin(
 
             level = rank.level
             tax_ids_to_keep = []
-            unclassified_tax_ids = []
+            unclassified_tax_ids = set()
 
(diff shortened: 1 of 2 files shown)


onecodex/analyses.py
```diff
@@ -367,7 +367,7 @@ def _to_classification_df(
 
             level = rank.level
             tax_ids_to_keep = []
-            unclassified_tax_ids = set()
```
jairideout, line 370: This needs to be a set because there could be duplicate tax IDs. To fix the pandas error, you can convert the set to a list on line 390 (the order of tax IDs doesn't matter because we're just summing their abundances)

jairideout (changes requested):

jairideout (approved):

PR #445, opened 2023-06-09
Handle samples with no abundances in heatmap

alenajk:
## Status
- [x] **Ready**

## Description

Currently, the custom plots abundances heatmap includes 0s even when the sample in question has no abundances calculated at all for it. For samples that have some abundances, but 0 for other taxa, a 0 value makes sense, but when a sample has no abundances at all, 0 should not be used as a stand-in. We would like to plot columns for samples fitting these criteria (no abundances calculated) as blank/none, and position them to one side of the chart.

image

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -42,10 +42,11 @@ class AnalysisMixin(
 
     def _get_auto_rank(self, rank):
         """Attempt to figure out what rank we should use for analyses."""
+        from onecodex.dataframes import OneCodexAccessor
 
         if rank == Rank.Auto:
             # if we're an accessor for a ClassificationsDataFrame, use its _rank property
-            if self.__class__.__name__ == "OneCodexAccessor":
+            if isinstance(self, OneCodexAccessor):
                 return self._rank
 
             if AbundanceMetric.has_value(self._metric) or self._is_metagenomic:
@@ -255,13 +256,23 @@ class AnalysisMixin(
         return magic_metadata, magic_fields
 
     @property
-    def all_nan_classification_ids(self):
-        if self.__class__.__name__ == "OneCodexAccessor":
-            if self._all_nan_classification_ids is None:
+    def _all_nan_classification_ids(self):
+        """
+        Provide list of classification ids for which there are no abundances calculated.
+
+        This can be used in plotting functions that may want to exclude or represent samples or
+        classifications in this category differently. Since plotting functions can be called on
+        `SampleCollection`s or `OneCodexAccessor`s, storing this list is safest, before
+        dataframe manipulation happens and information is possibly lost.
+        """
+        from onecodex.dataframes import OneCodexAccessor
+
+        if isinstance(self, OneCodexAccessor):
+            if self.ocx_all_nan_classification_ids is None:
                 # We rely on this list for accurate plot data, and it shouldn't ever be None, but if
                 # it is None, we raise an exception to avoid generating misleading plots
                 raise OneCodexException("Unable to fetch list of all_nan_classification_ids")
-            return self._all_nan_classification_ids
+            return self.ocx_all_nan_classification_ids
         return _get_all_nan_classification_ids(self._results)
 
     def to_df(self, analysis_type=AnalysisType.Classification, **kwargs):
@@ -451,7 +462,7 @@ class AnalysisMixin(
             "ocx_metric": self._metric,
             "ocx_taxonomy": self.taxonomy.copy(),
             "ocx_normalized": normalize,
-            "ocx_all_nan_classification_ids": self.all_nan_classification_ids,
+            "ocx_all_nan_classification_ids": self._all_nan_classification_ids,
         }
 
         # generate long-format table
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -19,6 +19,14 @@ from onecodex.viz import (
 )
 
 
+def get_all_nan_classification_ids(df):
+    all_nan_classification_ids = []
+    for class_id, is_all_nan in df.isnull().all(1).items():
+        if is_all_nan:
+            all_nan_classification_ids.append(class_id)
+    return all_nan_classification_ids
+
+
 class AnalysisMixin(
     VizPCAMixin, VizHeatmapMixin, VizMetadataMixin, VizDistanceMixin, VizBargraphMixin
 ):
@@ -245,6 +253,11 @@ class AnalysisMixin(
                         magic_fields[f] = str_f
 
         return magic_metadata, magic_fields
+
+    @property
+    def all_nan_classification_ids(self):
+        df = self._results.copy()
+        return get_all_nan_classification_ids(df)
 
     def to_df(self, analysis_type=AnalysisType.Classification, **kwargs):
         """
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -306,6 +306,7 @@ class AnalysisMixin(
         normalize="auto",
         table_format="wide",
         include_taxa_missing_rank=False,
+        include_nans=False,
     ):
         """Generate a ClassificationDataFrame, performing any specified transformations.
 
@@ -350,7 +351,7 @@ class AnalysisMixin(
                 )
 
         rank = self._get_auto_rank(rank)
-        df = self._results.copy()
+        df = self._collate_results(include_nans=include_nans).copy()
 
         # subset by taxa
         if rank:
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -351,7 +351,10 @@ class AnalysisMixin(
                 )
 
         rank = self._get_auto_rank(rank)
-        df = self._collate_results(include_nans=include_nans).copy()
+
+        df = self._results.copy()
+        if not include_nans:
+            df = df.fillna(0)
 
         # subset by taxa
         if rank:
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -256,8 +256,9 @@ class AnalysisMixin(
 
     @property
     def all_nan_classification_ids(self):
-        df = self._results.copy()
-        return get_all_nan_classification_ids(df)
+        if self.__class__.__name__ == "OneCodexAccessor":
+            return self._all_nan_classification_ids
+        return get_all_nan_classification_ids(self._results)
 
     def to_df(self, analysis_type=AnalysisType.Classification, **kwargs):
         """
@@ -321,7 +322,7 @@ class AnalysisMixin(
         include_taxa_missing_rank=False,
         include_nans=False,
     ):
-        """Generate a ClassificationDataFrame, performing any specified transformations.
+        """Generate a ClassificationsDataFrame, performing any specified transformations.
 
         Takes the ClassificationsDataFrame associated with these samples, or SampleCollection,
         does some filtering, and returns a ClassificationsDataFrame copy.
@@ -441,6 +442,7 @@ class AnalysisMixin(
             "ocx_metric": self._metric,
             "ocx_taxonomy": self.taxonomy.copy(),
             "ocx_normalized": normalize,
+            "_all_nan_classification_ids": self.all_nan_classification_ids,
         }
 
         # generate long-format table
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -19,7 +19,7 @@ from onecodex.viz import (
 )
 
 
-def get_all_nan_classification_ids(df):
+def _get_all_nan_classification_ids(df):
     all_nan_classification_ids = []
     for class_id, is_all_nan in df.isnull().all(1).items():
         if is_all_nan:
@@ -258,9 +258,11 @@ class AnalysisMixin(
     def all_nan_classification_ids(self):
         if self.__class__.__name__ == "OneCodexAccessor":
             if self._all_nan_classification_ids is None:
+                # We rely on this list for accurate plot data, and it shouldn't ever be None, but if
+                # it is None, we raise an exception to avoid generating misleading plots
                 raise OneCodexException("Unable to fetch list of all_nan_classification_ids")
             return self._all_nan_classification_ids
-        return get_all_nan_classification_ids(self._results)
+        return _get_all_nan_classification_ids(self._results)
 
     def to_df(self, analysis_type=AnalysisType.Classification, **kwargs):
         """
@@ -322,7 +324,8 @@ class AnalysisMixin(
         normalize="auto",
         table_format="wide",
         include_taxa_missing_rank=False,
-        include_nans=False,
+        fill_missing=True,
+        filler=0,
     ):
         """Generate a ClassificationsDataFrame, performing any specified transformations.
 
@@ -347,6 +350,10 @@ class AnalysisMixin(
         include_taxa_missing_rank : bool, optional
             Whether or not to include taxa that do not have a designated parent at `rank` (will be
             grouped into a "No <rank>" column).
+        fill_missing : bool, optional
+            Fill np.nan values
+        filler : float, optional
+            Value with which to fill np.nans
 
         Returns
         -------
@@ -369,8 +376,8 @@ class AnalysisMixin(
         rank = self._get_auto_rank(rank)
 
         df = self._results.copy()
-        if not include_nans:
-            df = df.fillna(0)
+        if fill_missing:
+            df = df.fillna(filler)
 
         # subset by taxa
         if rank:
@@ -444,7 +451,7 @@ class AnalysisMixin(
             "ocx_metric": self._metric,
             "ocx_taxonomy": self.taxonomy.copy(),
             "ocx_normalized": normalize,
-            "_all_nan_classification_ids": self.all_nan_classification_ids,
+            "ocx_all_nan_classification_ids": self.all_nan_classification_ids,
         }
 
         # generate long-format table
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -268,11 +268,11 @@ class AnalysisMixin(
         from onecodex.dataframes import OneCodexAccessor
 
         if isinstance(self, OneCodexAccessor):
-            if self.ocx_all_nan_classification_ids is None:
+            if self._ocx_all_nan_classification_ids is None:
                 # We rely on this list for accurate plot data, and it shouldn't ever be None, but if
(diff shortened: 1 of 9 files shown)


onecodex/models/collection.py
```diff
@@ -362,6 +368,8 @@ def _collate_results(self, metric=None, include_host=None):
         self._cached["results"] = df
```
audy, line 368: Suggestion: add a new function `_calculate_collated_results` that does all of the above math but doesn't save anything in cached results. Then, call said method in `_collate_results`. Use new method in `_cluster_by_sample` so that you avoid the cache altogether.

I think maybe we discussed this and realized that it wasn't that easy so apologies if this doesn't make sense.

lynchde:
Just testing locally with some of my own samples (abundances_w_children), and it looks as though we're clustering the samples without abundances with the samples that have 0 for the top_n species in the heatmap plot. I.e. it looks as though we're treating the no-abundance samples as 0% abundance in the clustering call.
(Screenshot without sample names. White bars are samples that don't have abundance estimates. Navy bars are samples that have abundance estimates, but the top 20 species for the collection are 0% for those samples). 
image

Is there any way we can force the non-abundance samples not to cluster amongst the 0% abundance for top_n samples? I.e. force the samples that don't have abundances to be on the right of the plot, and only cluster that samples that do have abundances estimated?

onecodex/dataframes.py
```diff
@@ -52,12 +59,14 @@ def __init__(
...
         ocx_metadata=None,
         ocx_normalized=None,
+        _all_nan_classification_ids=None,
     ):
         self.ocx_rank = ocx_rank
         self.ocx_metric = ocx_metric
         self.ocx_taxonomy = ocx_taxonomy
         self.ocx_metadata = ocx_metadata
         self.ocx_normalized = ocx_normalized
+        self._all_nan_classification_ids = _all_nan_classification_ids
```
audy, line 69: prepend `ocx_` instead of `_`

onecodex/analyses.py
```diff
@@ -306,8 +322,9 @@ def _to_classification_df(
         normalize="auto",
         table_format="wide",
         include_taxa_missing_rank=False,
+        include_nans=False,
```
audy, line 325: Make kwarg names consistent with `_to_functional_df` (`filler` & `fill_missing`). I think the default will have to be different though for backwards compatibility.

onecodex/distance.py
```diff
@@ -141,7 +141,7 @@ def unifrac(self, weighted=True, rank=Rank.Auto):
         """
         import skbio.diversity
 
-        df = self.to_df(rank=rank, normalize=self._guess_normalized())
+        df = self.to_df(rank=rank, normalize=self._guess_normalized()).fillna(0)
```
audy, line 144: You should be able to pass `replace_nans` and `filler` to `to_df`

onecodex/models/collection.py
```diff
@@ -350,7 +353,10 @@ def _collate_results(self, metric=None, include_host=None):
                 if not include_host and d_tax_id in host_tax_ids:
                     continue
 
-                data[c_idx, tax_id_to_idx[d_tax_id]] = d[metric] or 0
+                if metric in [Metric.AbundanceWChildren, Metric.Abundance]:
```
audy, line 356: I don't think you need this check anymore after lines 342-343. The default value should already be `0` for metrics other than `AbundanceWChildren` and `Abundance`

onecodex/analyses.py
```diff
@@ -246,6 +254,14 @@ def _metadata_fetch(self, metadata_fields, label=None):
 
         return magic_metadata, magic_fields
 
+    @property
+    def all_nan_classification_ids(self):
+        if self.__class__.__name__ == "OneCodexAccessor":
+            if self._all_nan_classification_ids is None:
```
audy, line 260: Add comment about why this check is necessary

onecodex/analyses.py
```diff
@@ -19,6 +19,14 @@
 )
 
 
+def get_all_nan_classification_ids(df):
```
audy, line 22: Prepend with `_` since this isn't being used anywhere else

onecodex/viz/_distance.py
```diff
@@ -42,19 +42,33 @@ def _compute_distance(self, rank, metric):
         return distances
 
     def _cluster_by_sample(
-        self, rank=Rank.Auto, metric=BetaDiversityMetric.BrayCurtis, linkage=Linkage.Average
+        self,
+        all_nan_classification_ids=None,
```
audy, line 46: I thin you can get this from `self.all_nan_classification_ids` since that's defined on `AnalysisMixin` which works for both `ClassificationsDataFrame` and `SampleCollection`

onecodex/analyses.py
```diff
@@ -246,6 +254,14 @@ def _metadata_fetch(self, metadata_fields, label=None):
 
         return magic_metadata, magic_fields
 
+    @property
+    def all_nan_classification_ids(self):
+        if self.__class__.__name__ == "OneCodexAccessor":
```
audy, line 259: Add a comment that this case happens when `self` is a `ClassificationsDataFrame`

onecodex/viz/_heatmap.py
```diff
@@ -238,6 +251,25 @@ def plot_heatmap(
...
+            if self.__class__.__name__ == "OneCodexAccessor":
+                # the df of a `OneCodexAccessor` does not have NaNs, and we want to display
+                # all_nan_classificaion_id columns as white instead of color corresponding to 0
+                d = d.replace(0.0, np.nan)
+            dropped.append(d)
+            index = df[(df == classification_id).any(axis=1)].index
+            df = df.drop(index)
+
+        df = df.replace(np.nan, 0)
+        for dropped_df in dropped:
```
audy, line 268: optimization (`pd.concat` is slow): `pd.concat([df, *dropped])`

audy (changes requested):

onecodex/dataframes.py
```diff
@@ -52,12 +59,14 @@ def __init__(
         ocx_taxonomy=None,
         ocx_metadata=None,
         ocx_normalized=None,
+        _all_nan_classification_ids=None,
```
audy, line 62: prepend ocx_ instead of _ for consistency

onecodex/models/collection.py
alenajk: The value in d[metric] would still be possibly `None` though, so as far as I can see we either need this check or to replace what we have with something like
```
if d[metric]:
    data[c_idx, tax_id_to_idx[d_tax_id]] = d[metric]
```
Maybe that's clearer ^ - wdyt?

audy: oops. You're right. I think `d[metric] or 0` is clearer.

onecodex/viz/_distance.py
alenajk: True, but this is used by `plot_distance`, in which we actually want to exclude the samples/classifications entirely from the plot (in a separate PR). Since the behavior is different, and because we actually use more than just the `ids_in_order`, I wanted to keep this as a parameter so that `plot_distance` could _not_ pass anything and keep its current behavior for now. Let me know if you have thoughts though or another idea about handling these two cases.

audy (approved):

GSGerritsen (approved):
This seems like a very reasonable way to handle the nan stuff. One suggestion is that it could be nice to put the text from the PR description (or some variant of it) _somewhere_ in the code. It serves as a nice explanation for all the special handling with nans, up to you though!

onecodex/viz/_heatmap.py
```diff
@@ -238,6 +251,24 @@ def plot_heatmap(
...
+                # all_nan_classification_id columns as white instead of color corresponding to 0
+                d = d.replace(0.0, np.nan)
+            dropped.append(d)
+            index = df[(df == classification_id).any(axis=1)].index
+            df = df.drop(index)
+
+        df = df.replace(np.nan, 0)
+        df = pd.concat([df, *dropped])
+
+        assert set(df["Label"].values) == set(labels_in_order)
```
jairideout, line 270: is this assert needed?

onecodex/viz/_heatmap.py
```diff
@@ -238,6 +251,24 @@ def plot_heatmap(
                 haxis, header=alt.Header(titleOrient="bottom", labelOrient="bottom")
             )
 
+        # Drop all-NaN rows, convert remaining NaNs to 0s, and concat
+        # dropped all-NaN rows to the end
+        dropped = []
+        for classification_id in all_nan_classification_ids:
```
jairideout, line 257: Since we know which rows we want to move to the end, I think we can accomplish this with less work:

```python
ids_with_data = []
for classification_id in df.index:
    if classification_id not in all_nan_classification_ids:
        ids_with_data.append(classification_id)

df = df.reindex(ids_with_data + all_nan_classification_ids)
```

onecodex/analyses.py
```diff
@@ -246,6 +254,16 @@ def _metadata_fetch(self, metadata_fields, label=None):
 
         return magic_metadata, magic_fields
 
+    @property
+    def all_nan_classification_ids(self):
```
jairideout, line 258: Can we make this a "private" property, i.e. `_all_nan_classification_ids`? It's a pretty internal thing that we might not want to commit to having in the public API (yet)

tests/test_viz.py
```diff
@@ -254,6 +254,80 @@ def test_plot_heatmap(ocx, api_data):
...
+        .isnull()
+        .any()
+        == False  # noqa
+    )
+
+    # Sample with all NaNs should be at the end
+    assert chart.encoding.x.sort[-1] == all_nan_sample.filename
+
+
+def test_plot_heatmap_on_df_plots_all_nan_samples_with_nans(ocx, api_data):
```
jairideout, line 294: what's the difference between this test and the one above it?

onecodex/models/collection.py
```diff
@@ -339,6 +339,9 @@ def _collate_results(self, metric=None, include_host=None):
 
         data = np.zeros((len(self._classifications), len(tax_info.index)), dtype=metric_dtype)
 
+        if metric in [Metric.AbundanceWChildren, Metric.Abundance]:
+            data[:] = np.nan
```
jairideout, line 343: `ndarray.fill()` is a bit faster ([link](https://stackoverflow.com/a/1704853/3776794)):

```python
data.fill(np.nan)
```

The only reason I mention it is because this part of the code is the slowest part when plotting lots of data

onecodex/viz/_heatmap.py
alenajk: Not strictly - @audy had suggested it to make sure that we aren't ever ending up in a situation where we have a mismatch between the chart and the `labels_in_order`, which would lead to an incorrect plot

tests/test_viz.py
alenajk: This tests that the functionality works for `plot_heatmap` called on a `OneCodexAccesor` , ie `samples.to_df().ocx.plot_heatmap`. Since `OneCodexAccessor` uses its own version of `_results`, we need to make sure that we're getting the same functionality (by storing `all_nan_classification_ids` on both `SampleCollection`s and `OneCodexAccessor`s) in both cases

onecodex/analyses.py
alenajk: Sure, that makes sense (although it will necessitate renaming the attribute on `OneCodexAccessor`s something else besides `_all_nan_classification_ids` to distinguish between the two). But also, how does leaving it as-is commit us to anything?

onecodex/viz/_heatmap.py
alenajk: We also need to fill the df with 0s for the `None`s that _don't_ correspond to classifications with no abundances calculated. Also, I think we would still need to replace `0.0` with `np.nan` in the case that we are dealing with a `OneCodexAccessor`

onecodex/models/collection.py
```diff
@@ -350,7 +353,10 @@ def _collate_results(self, metric=None, include_host=None):
                 if not include_host and d_tax_id in host_tax_ids:
                     continue
 
-                data[c_idx, tax_id_to_idx[d_tax_id]] = d[metric] or 0
+                if metric in [Metric.AbundanceWChildren, Metric.Abundance]:
+                    data[c_idx, tax_id_to_idx[d_tax_id]] = d[metric]
```
pkonieczny, line 357: Should we use `d[metric] or np.nan` here?

onecodex/viz/_distance.py
```diff
@@ -64,14 +79,18 @@ def _cluster_by_sample(
...
+        import numpy as np
         from scipy.cluster import hierarchy
         from scipy.spatial.distance import squareform
         from sklearn.metrics.pairwise import euclidean_distances
 
-        dist_matrix = euclidean_distances(self._results.T).round(6)
-        clustering = hierarchy.linkage(squareform(dist_matrix), method=linkage)
+        df = self._results.dropna(how="all").replace(np.nan, 0)
+
+        dist_matrix = euclidean_distances(df.T).round(6)
```
pkonieczny, line 89: Could we use https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.nan_euclidean_distances.html#sklearn.metrics.pairwise.nan_euclidean_distances and not fill in 0's here?

pkonieczny (approved):
I'm not a big fan of `all_nan_classification_id` params, but I do think it's better to merge as is and refactor later on.

onecodex/viz/_distance.py
pkonieczny: Scratch that, we do want the distances to 0 calculated

onecodex/viz/_heatmap.py
jairideout: I'm confused why a classification in `all_nan_classification_ids` has zeros instead of nans (the "all nan" part of the name is kind of misleading). Is there a way to handle that upstream from here so that `all_nan_classification_ids` will always point to classifications that are all nans?

onecodex/viz/_heatmap.py
```diff
@@ -238,6 +251,24 @@ def plot_heatmap(
...
+            d = df[(df == classification_id).any(axis=1)]
```
jairideout, line 258: To get a specific row in a dataframe by index label:

```python
row = df.loc[classification_id]
```

onecodex/viz/_heatmap.py
```diff
@@ -238,6 +251,24 @@ def plot_heatmap(
...
+            if self.__class__.__name__ == "OneCodexAccessor":
+                # the df of a `OneCodexAccessor` does not have NaNs, and we want to display
+                # all_nan_classification_id columns as white instead of color corresponding to 0
+                d = d.replace(0.0, np.nan)
+            dropped.append(d)
+            index = df[(df == classification_id).any(axis=1)].index
```
jairideout, line 264: lines 264 and 265 can be replaced with:

```python
df = df.drop(classification_id)
```

jairideout:
@alenajk @audy maybe it'd be simpler to filter out all nan / zeros samples in `to_df()`, like we do for taxa with the `remove_zeros=True` param? We could raise a warning if any samples are removed, and Custom Plots already catches `PlottingWarning`s and displays them to the user. That's the behavior we want for the beta diversity distance plots ([DEV-8400](https://linear.app/onecodex/issue/DEV-8400/beta-diversity-distance-plot-should-not-cluster-samples-that-we-dont)), maybe we could do the same for heatmaps too?

onecodex/models/collection.py
alenajk: I think it shouldn't actually make a difference, but for consistency it is probably better

onecodex/viz/_heatmap.py
alenajk: I'm getting `KeyError` with that

onecodex/viz/_heatmap.py
alenajk: This isn't working for me, actually. It leads to a `KeyError`

alenajk:
@jairideout Thanks for the review! Re: `all_nan_classification_ids` sometimes having 0s, it's not really under our control, since even though we have paramerized `to_df` to allow for _not_ filling with 0s, you can call `plot_heatmap` on a `OneCodexAccesor`, and when it calls `to_df` internally, `to_df` references `self._results`, which is just the df you've instantiated the `OneCodexAccessor` with. This also is why we went further back instead of doing filtering in `to_df` or `to_classification_df`.

I can rename for clarity though - perhaps `no_abundances_classification_ids` or `classification_ids_with_no_abundances` would be clearer.

onecodex/models/collection.py
jairideout: Hmm but then `0.0` or `0` abundances will get coerced into `np.nan`. Is that something we want?

onecodex/viz/_heatmap.py
jairideout: oh my bad, I thought the rows in the dataframe were classification IDs, and the columns were taxa, but that's only the case when `to_df(table_format="wide")` is used (the format is `"long"` here)

onecodex/viz/_heatmap.py
jairideout: whoops, didn't realize we were using a `long` format df instead of `wide`

onecodex/analyses.py
alenajk: I have renamed it

onecodex/models/collection.py
alenajk: Ah, good point - no, we want to keep 0 and 0.0. I will revert to d[metric] in this case

onecodex/dataframes.py
```diff
@@ -172,7 +172,7 @@ def __init__(self, pandas_obj):
         self._metric = pandas_obj.ocx_metric
         self._rank = pandas_obj.ocx_rank
         self._results = pandas_obj
-        self._all_nan_classification_ids = pandas_obj.ocx_all_nan_classification_ids
+        self.ocx_all_nan_classification_ids = pandas_obj.ocx_all_nan_classification_ids
```
jairideout, line 175: the other attributes on this object aren't prefixed with `ocx_`, is there a reason this one is? the previous name `_all_nan_classification_ids` made sense to me, and it's a "private" name too

alenajk: It can't be the same name as the property on `AnalysisMixin` - I can put an underscore though

jairideout (approved):

PR #452, opened 2023-07-22
Fix plotting faceted heatmaps w/o abundances

audy:
## Status
- [x] **Ready**

## Description

* Issue is related to how we're subsetting the internal/results data frame when faceting. 

## Related PRs
- [x] This PR is independent

## TODOs

- [x] Tests


Keats (approved):

tests/test_viz.py
```diff
@@ -303,6 +303,24 @@ def test_plot_heatmap_plots_all_nan_samples_with_nans(ocx, api_data, is_onecodex
     assert chart.encoding.x.sort[-1] == all_nan_sample.filename
 
 
+def test_plot_heatmap_with_haxis_two_cluster_groups(ocx, api_data):
+    import numpy as np
+
+    samples = ocx.Samples.where(project="4b53797444f846c4")
+    sample1 = ocx.Samples.get("cc18208d98ad48b3")
```
jairideout, line 310: Do we actually need all that new test data? We could probably set one of the existing sample's abundances in that project to `np.nan` to reproduce the bug

audy: @jairideout there are branches in the code that only execute if there are >3 samples so we had to add some

tests/test_viz.py
```diff
@@ -303,6 +303,42 @@ def test_plot_heatmap_plots_all_nan_samples_with_nans(ocx, api_data, is_onecodex
...
+    # Set abundances for one sample to be all NaN
+    # (It should not be in the clustering group)
+    all_nan_sample = samples[1]
+    samples._results.loc[all_nan_sample.primary_classification.id] = np.nan
+    assert len(samples._all_nan_classification_ids) == 1
+
+    # # Make sure we are actually calling _cluster_by_sample
+    with mock.patch(
+        "onecodex.viz._distance.VizDistanceMixin._cluster_by_sample",
+        return_value="['4eed25415f6945dd', '6579e99943f84ad2', 'e0422602de41479f', '1198c7ce565643f9']",
```
jairideout, line 323: why do we need to mock the return_value in two places?

jairideout (approved):

tests/test_viz.py
alenajk: Ah, it's not, was just trying different ways -- updated!

PR #453, opened 2023-07-22
Fix CI for Python 3.7

audy:
## Status
- [x] **Ready**

## Description

Cython 3 breaks scikit-bio installation under Python 3.7 (scikit-bio>=0.5.8 has other incompatibility issues with Python 3.7). Since these are [build dependencies](https://github.com/biocore/scikit-bio/blob/0.5.7/pyproject.toml#L2), they cannot be pinned in `setup.py`.

This PR manually installs build deps for scikit-bio==0.5.7 under Python 3.7 to pin the Cython version.

## Related PRs
- [x] This PR is independent

## TODOs

- [x] Tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,12 @@ jobs:
         run: python -c "import sys; print(sys.version)"
       - name: Install dependencies in a venv
         run: |
+          # fix for installing scikit-bio==0.5.7 on python-3.7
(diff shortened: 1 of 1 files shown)


Keats:
3.7 is EOL, should we just drop supporting it?

jairideout (approved):

jairideout:
> 3.7 is EOL, should we just drop supporting it?

We need to find out if any of our customers are using py3.7, otherwise we can drop it

PR #455, opened 2023-07-28
exclude all-nan samples from distance heatmap

alenajk:
## Status
- [x] **Ready**

## Description
We want to totally exclude samples that have no abundances calculated for them from the distance heatmap plot. This adds `exclude_all_nan` to a few places with a default of `False`, and a call to `_cluster_by_sample` with `exclude_all_nan=True` to get the desired results.

## Related PRs
- [ x] This PR is independent

## TODOs

- [x] Tests

### DIFF
--- a/onecodex/viz/_distance.py
+++ b/onecodex/viz/_distance.py
@@ -208,10 +208,9 @@ class VizDistanceMixin(DistanceMixin):
 
         if self._all_nan_classification_ids:
             warnings.warn(
-                PlottingWarning(
-                    f"{len(self._all_nan_classification_ids)} sample(s) have no abundances "
-                    "calculated and have been omitted from the distance heatmap."
-                )
+                f"{len(self._all_nan_classification_ids)} sample(s) have no abundances "
+                "calculated and have been omitted from the distance heatmap.",
+                PlottingWarning,
             )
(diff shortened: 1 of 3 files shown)


jairideout (changes requested):

onecodex/viz/_distance.py
```diff
@@ -185,7 +198,21 @@ def plot_distance(
...
+        clust = self._cluster_by_sample(
+            rank=rank,
+            metric=metric,
+            linkage=linkage,
+            all_nan_classification_ids=self._all_nan_classification_ids,
+            exclude_all_nan=True,
+        )
+
+        if self._all_nan_classification_ids:
+            warnings.warn(
```
jairideout, line 210: This doesn't raise a warning for me. The first arg to `warnings.warn` should be the message and the second arg is the warning type:

```python
warnings.warn(
    f"{len(self._all_nan_classification_ids)} sample(s) have no abundances calculated and have been omitted from the distance heatmap.",
    PlottingWarning,
)
```

jairideout (changes requested):

jairideout (approved):

audy (approved):

PR #461, opened 2023-08-31
Add optional name to asset upload

alenajk:
## Status
- [x] **Ready**

## Description
Add a `--name` option to `asset upload` command. This can be helpful so that a user can give a file/asset a more user-friendly name, rather than it using the filename by defaul.

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -247,6 +247,7 @@ def assets():
     "--name",
     nargs=1,
     required=False,
+    help=OPTION_HELP["name"],
 )
 @click.argument(
     "file",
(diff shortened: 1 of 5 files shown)


onecodex/cli.py
```diff
@@ -264,7 +269,7 @@ def assets_upload(ctx, max_threads, file):
     run_via_threadpool(
         ctx.obj["API"].Assets.upload,
         [file],
-        {"progressbar": bar},
+        {"progressbar": bar, "name": name},
```
Keats, line 272: should we validate that the name is not empty?

Keats (approved):

onecodex/cli.py
alenajk: The `name` option is required so it won't work if nothing is passed, but passing an empty string is permitted so I can add a guard for that case

jairideout (approved):

pkonieczny (approved):
Looks good!

PR #466, opened 2023-09-18
Refactor metadata, add FunctionalDataFrame

pkonieczny:
## Status
- [X] **Ready**

## Description
This PR refactors the metadata preparation code so that it will work with functional profiles without fetching the classification results as well.
The functional result dataframe is now indexed with sample id instead of functional profile id so that it may be connected with the corresponding sample metadata.
Also in this PR: introducing `FunctionalDataFrame` class which encapsulates functional result pandas dataframes.

## Related PRs
- [X] This PR is independent


onecodex/analyses.py
```diff
@@ -307,7 +314,7 @@ def _to_functional_df(
         filler=0,
     ):
         """
-        Return a dataframe of results from a functional analysis.
+        Return a `ClassificationsDataFrame` of results from a functional analysis.
```
petraszd, line 317: Nitpick: It returns `FunctionalDataFrame`.

petraszd (approved):

PR #467, opened 2023-09-18
Add functional results feature name map to df

pkonieczny:
## Status
- [X] **Ready**

## Description
Handle Functional Analysis results feature id -> name map by adding it as additional metadata to functional df.

## Related PRs
- [X] This PR is independent

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -333,7 +333,7 @@ class AnalysisMixin(
         """
         from onecodex.dataframes import FunctionalDataFrame
 
-        df = self._functional_results(
+        df, feature_name_map = self._functional_results(
             annotation=annotation,
             taxa_stratified=taxa_stratified,
             metric=metric,
@@ -345,6 +345,7 @@ class AnalysisMixin(
             ocx_metadata=self.metadata.copy(),
             ocx_functional_group=annotation,
             ocx_metric=metric,
+            ocx_feature_name_map=feature_name_map,
         )
 
     def _to_classification_df(
--- a/onecodex/dataframes.py
+++ b/onecodex/dataframes.py
@@ -151,7 +151,7 @@ class ClassificationsSeries(pd.Series):
 class FunctionalDataFrame(pd.DataFrame):
     """A DataFrame containing additional One Codex metadata."""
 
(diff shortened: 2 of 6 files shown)


onecodex/analyses.py
```diff
@@ -333,7 +333,7 @@ def _to_functional_df(
         """
         from onecodex.dataframes import FunctionalDataFrame
 
-        df = self._functional_results(
+        df, feature_name_map = self._functional_results(
```
petraszd, line 336: Why it is better than embedding it into `df` as an additional column?

petraszd: Pandas do support indexing rows by tuples. I am not sure about implications to performance though (@lynchde ?).

```python
import pandas as pd

df = pd.DataFrame({
    "a": [1, 2, 3],
    "b": [4, 5, 6],
    "c": [7, 8, 9],
}, index=[uuid_helper(), uuid_helper(), uuid_helper()])

print(df)
#                   a  b  c
# 7b332a8081444895  1  4  7
# 7a9918b9c45b4506  2  5  8
# d52dd70806264c7d  3  6  9

new_index = list(zip(df.index, ["X", "Y", "Z"]))
df.index = pd.Index(new_index)

print(df)
#                     a  b  c
# 7b332a8081444895 X  1  4  7
# 7a9918b9c45b4506 Y  2  5  8
# d52dd70806264c7d Z  3  6  9

idx0 = df.index[0]
print(type(idx0), idx0)
# <class 'tuple'> ('7b332a8081444895', 'X')
```

pkonieczny: That's mostly for consistency, classification results and their additional data are handled in the same way. 
The df is functional profiles (or samples) x feature values, embedding feature names would require either a 3d matrix or additional hacks that would drop the name column when not needed 🤷

petraszd: It is not a multicolumn index. It is just a tuple.

petraszd: Let's leave it with magix `ocx_MAGIC` attributres. For consistency.

petraszd (approved):

onecodex/analyses.py
pkonieczny: Right - but the issue with joining with metadata df still stands, we don't have feature names there.

PR #469, opened 2023-09-19
Fixing sorting and joining

petraszd:
## Status
- [x] **Ready**

## Description

- `reindex` does not work that way.
- Replacing `function_id` with names messes up sorting.

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/viz/_functional.py
+++ b/onecodex/viz/_functional.py
@@ -1,5 +1,3 @@
-import pandas as pd
-
 from onecodex.lib.enums import (
     FunctionalAnnotations,
     FunctionalAnnotationsMetric,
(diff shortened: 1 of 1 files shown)


onecodex/viz/_functional.py
```diff
@@ -67,40 +69,38 @@ def plot_functional_heatmap(
...
 
-        function_id_to_name = df.ocx_feature_name_map
-
         df.drop(columns=to_drop.index, inplace=True)
 
+        # TODO: comment to explain
         metadata = df.ocx_metadata
         if metadata.index.name != "sample_id":
-            metadata = metadata.reindex(metadata["sample_id"])
+            metadata.index = pd.Index(metadata["sample_id"], name="sample_id")
```
pkonieczny, line 77: Maybe we should call `set_index`?

pkonieczny (approved):

onecodex/viz/_functional.py
```diff
@@ -1,3 +1,5 @@
+import pandas as pd
```
pkonieczny, line 1: Let's defer loading that to inside the function?

PR #471, opened 2023-09-20
Functional Heatmap: Sorting and labeling

petraszd:
## Status
- [x] **Ready**

## Description

Adds ability to label by and sort by x-axis

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -1,4 +1,3 @@
-import six
 import warnings
 from collections import Counter
 
@@ -520,7 +519,7 @@ class AnalysisMixin(
         elif callable(label):
             for item_id, item_meta in metadata.to_dict(orient="index").items():
                 item_label = label(item_meta)
-                if not isinstance(item_label, six.string_types):
+                if not isinstance(item_label, str):
                     wrong_type = type(item_label).__name__
                     raise OneCodexException(
                         "Expected string from label function, got: {}".format(wrong_type)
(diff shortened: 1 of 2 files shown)


onecodex/analyses.py
```diff
@@ -536,6 +490,63 @@ def _get_highest_unclassified_tax_id(self, tax_id, level):
...
+            else:
+                raise OneCodexException(
+                    "Label field {} not found. Choose from: {}".format(
+                        label, ", ".join(metadata.keys())
+                    )
+                )
+        elif callable(label):
+            for item_id, item_meta in metadata.to_dict(orient="index").items():
+                item_label = label(item_meta)
+                if not isinstance(item_label, six.string_types):
```
pkonieczny, line 523: Since `six.string_types` was already replaced with `str` a few lines up, should we do the same here?

onecodex/analyses.py
```diff
@@ -536,6 +490,63 @@ def _get_highest_unclassified_tax_id(self, tax_id, level):
...
+                raw_result[item_id] = item_label
+
+        elif label is not None:
+            wrong_type = type(label).__name__
+            raise OneCodexException(
+                "Expected string or callable for label, got: {}".format(wrong_type)
+            )
+
+        # add an incremented number to duplicate labels (e.g., same filename)
+        duplicates_counter = Counter(raw_result.values())
```
pkonieczny, line 537: Is the counter really faster than pandas `duplicated` calls? Also, should we drop `na` values first?

petraszd: Python dict speed would become and issue with 10K items. So I do not worry about the speed. `na` values are valid comment. I will see maybe I can convert dict to `pd.Series` ealier.

petraszd: Turns out dropping `na` was just an artifact of calling `where` and `duplicated`.

PR #473, opened 2023-09-21
update handling of missing phenotype

alenajk:
## Status
- [x] **Ready**

## Description
If samples don't have a specified `haxis`, it breaks plotting. With this change, we update the `0`s to `None`s so that those samples will be clustered by themselves with the `None` label.

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/viz/_heatmap.py
+++ b/onecodex/viz/_heatmap.py
@@ -152,6 +152,9 @@ class VizHeatmapMixin(object):
 
         # and for metadata
         for f in tooltip:
+            if f == haxis:
+                magic_metadata[[haxis]] = magic_metadata[[haxis]].replace(np.nan, "None")
+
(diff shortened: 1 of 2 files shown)


audy (approved):
LGTM

onecodex/viz/_heatmap.py
```diff
@@ -145,6 +145,7 @@ def plot_heatmap(
         tooltip.insert(0, "Label")
 
         magic_metadata, magic_fields = self._metadata_fetch(tooltip, label=label)
+        magic_metadata.replace(np.nan, "None", inplace=True)
```
jairideout, line 148: Can we use `"N/A"` to match the other places where we display missing values?

Also a unit test would be nice

alenajk: Updated

jairideout (approved):

PR #477, opened 2023-09-27
Select the newest Functional Profiles for sample collections

pkonieczny:
## Status
- [X] **Ready**

## Description
~~This PR adds Functional Profile filtering: when fetching Functional Profiles for a collection of samples, the newest run job is determined first, then the Functional Profiles are filtered so that only ones corresponding to the newest job remain.~~
This PR adds Functional Profile version mixing. If a sample has more than one Functional Profile, the run with newer job is selected. If there are more runs for the same job (highly unlikely and probably an error, but possible), the more recent run is selected.
If more than a single Functional Analysis version is encountered, a proper warning message will be shown.

Also in this PR: fixes for the magic potion logic around `$in` where conditions.

## Related PRs
- [X] This PR is independent


onecodex/models/collection.py
```diff
@@ -417,54 +420,43 @@ def _functional_profiles_fetch(self, skip_missing=None):
...
+        profiles = []
+        for i in range(0, len(sample_ids), batch_size):
+            profiles += FunctionalProfiles.where(sample=sample_ids[i : i + batch_size])
+        if not profiles:
+            self._cached["functional_profiles"] = []
+            return
+
+        # Determine the newest Functional Analysis job for this sample set
+        job_ids = {fp.job.id for fp in profiles}
+        jobs = Jobs.where(id=sorted(job_ids), sort="created_at")
```
petraszd, line 441: Nitpick: what is the reason to sort ids?

onecodex/models/collection.py
```diff
@@ -417,54 +420,43 @@ def _functional_profiles_fetch(self, skip_missing=None):
...
-        # ensure all the functional profiles are the same job version
-        job_ids = set([obj.job.id for obj in functional_profiles])
-        if len(job_ids) > 1:
-            raise OneCodexException(
-                "FunctionalProfiles contain multiple analysis job versions: {}".format(
-                    ", ".join(job_ids)
-                )
-            )
+                    raise OneCodexException(
+                        f"Current version of functional profile not found for sample {sample_id}."
```
petraszd, line 456: Nitpick: strickly speaking there is a theoretical possibility that `newest_job` is not a current job.

petraszd (approved):

onecodex/models/collection.py
pkonieczny: I've changed the text, but I don't think it will help the user at all - and I don't think it's worth explaining the whole process of determining the "set of functional profiles for the newest job for this particular collection" in an error message 🤷

onecodex/models/collection.py
```diff
@@ -408,6 +408,9 @@ def taxonomy(self):
     def _functional_profiles_fetch(self, skip_missing=None):
         """Transform a list of Samples or Classifications into a list of FunctionalProfiles objects.
 
+        The FunctionalProfiles are filtered down to only the ones matching the newest job version
```
petraszd, line 411: This is not true anymore.

petraszd (approved):

onecodex/models/collection.py
pkonieczny: Haha just noticed and pushed an update

PR #478, opened 2023-10-16
Minor functional results improvements

pkonieczny:
## Status
- [X] **Ready**

## Description
This PR includes minor performance improvements for the functional result dataframe forming which should benefit when plotting functional run results.

## Related PRs
- [X] This PR is independent

### DIFF
--- a/onecodex/api.py
+++ b/onecodex/api.py
@@ -107,7 +107,7 @@ class Api(object):
 
         # Set backoff / retry strategy for 429s
         retry_strategy = Retry(
-            total=3, backoff_factor=4, allowed_methods=False, status_forcelist=[429, 502, 503]
+            total=3, backoff_factor=4, allowed_methods=None, status_forcelist=[429, 502, 503]
         )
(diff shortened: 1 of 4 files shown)


onecodex/api.py
```diff
@@ -107,7 +107,7 @@ def __init__(
 
         # Set backoff / retry strategy for 429s
         retry_strategy = Retry(
-            total=3, backoff_factor=4, allowed_methods=False, status_forcelist=[429, 502, 503]
+            total=3, backoff_factor=4, allowed_methods=None, status_forcelist=[429, 502, 503]
```
pkonieczny, line 110: `allowed_methods` is not a boolean param

jairideout: In urllib3 < v2.0 their docs recommend passing `False` to allow all methods. In v2.0 their docs recommend passing `None`. Since this package depends on urllib3 < v2.0, we might want to pass `allowed_methods=False`. I haven't tested to see if `None` works, it's possible any falsy value will be accepted.

PR #481, opened 2023-11-27
Func Viz: using same axis definitions as in other viz functions

petraszd:
## Status
- [x] **Ready**

## Description

Altair allows you to pass `title` directly without wrapping it with `Axis`. But all other visualisation functions are using `Axis` constructor making functional plot a special case.

This PR makes it act the same as others.

## Related PRs
- [x] This PR is independent


pkonieczny (approved):
LGTM 👍

PR #485, opened 2024-01-10
Fix date field type coercion in plot_metadata() & add `coerce_haxis_dates` parameter

jairideout:
## Status
- [x] **Ready**

## Description
In `plot_metadata()`, column names containing the word "date" (after splitting on underscores) have their column dtype coerced to datetime. This didn't work correctly because calling `chart.to_dict()` would raise a JSON serialization `TypeError`.

Fixed the type coercion and added a `coerce_haxis_dates` parameter to control whether type coercion happens (the default is `True` for backwards-compatibility).

Closes DEV-9026

## Related PRs
- [x] This PR is independent

### DIFF
--- a/tests/test_viz.py
+++ b/tests/test_viz.py
@@ -8,6 +8,7 @@ pytest.importorskip("pandas")  # noqa
 
 import altair as alt
 import numpy as np
+import pandas as pd
 
 from onecodex.lib.enums import Metric, AbundanceMetric, Link, Rank, BetaDiversityMetric, Linkage
 from onecodex.exceptions import OneCodexException, PlottingException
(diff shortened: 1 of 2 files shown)


tests/test_viz.py
```diff
@@ -101,6 +102,24 @@ def test_plot_metadata_alpha_diversity_with_nans(ocx, api_data):
...
+def test_plot_metadata_haxis_date(ocx, api_data, coerce_haxis_dates):
+    samples = ocx.Samples.where(project="4b53797444f846c4")
+    samples.metadata["date_collected"] = datetime.now().isoformat()
+
+    chart = samples.plot_metadata(
+        vaxis="shannon",
+        haxis="date_collected",
+        return_chart=True,
+        coerce_haxis_dates=coerce_haxis_dates,
+    )
```
audy, line 115: it would be good to add a check that the data did or did not get coerced here.

audy (changes requested):

audy (approved):

PR #490, opened 2024-02-14
Add secondary haxis grouping to `plot_metadata()`

jairideout:
## Status
- [x] **Ready**

## Description
Added optional `secondary_haxis` parameter to `plot_metadata()`, which can be a field name or tuple of field names.

If provided, a grouped scatterplot or boxplot is generated, where `haxis` is the primary grouping and `secondary_haxis` is the secondary grouping. The secondary grouping is colored and includes a legend.

Closes DEV-9216

In the example below, `haxis="Cage"` and `secondary_haxis="Cohort"`:

Screenshot 2024-02-14 at 2 23 43 PM

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/viz/_metadata.py
+++ b/onecodex/viz/_metadata.py
@@ -30,6 +30,7 @@ class VizMetadataMixin(object):
         height=400,
         facet_by=None,
         coerce_haxis_dates=True,
+        secondary_haxis=None,
     ):
         """Plot an arbitrary metadata field versus an arbitrary quantity as a boxplot or scatter plot.
 
@@ -81,6 +82,10 @@ class VizMetadataMixin(object):
             underscores) will be coerced to datetime dtype. For example, the field "date_collected"
             will be coerced if ``coerce_haxis_dates`` is ``True``.
 
+        secondary_haxis : str or tuple of str, optional
+            The secondary metadata field (or tuple containing multiple categorical fields) to be
+            plotted on the horizontal axis.
+
         Examples
         --------
         Generate a boxplot of the abundance of Bacteroides (genus) of samples grouped by whether the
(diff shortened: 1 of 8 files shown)


onecodex/viz/_metadata.py
```diff
@@ -171,35 +192,43 @@ def plot_metadata(
...
+            encode_kwargs["column"] = alt.Column(
                 facet_by, header=alt.Header(titleOrient="bottom", labelOrient="bottom")
             )
 
+        x_kwargs = {"axis": alt.Axis(title=xlabel)}
+        if secondary_haxis:
+            x_kwargs.update(
+                {
+                    "title": None,
+                    "axis": alt.Axis(title=xlabel, labels=False, ticks=False),
```
pkonieczny, line 206: Why are we setting `labels` and `ticks` explicitly to `False` here and not above?

tests/test_viz.py
```diff
@@ -87,6 +87,34 @@ def test_plot_metadata_facet_by_boxplot(ocx, api_data):
     assert chart.encoding.x.axis.title == ""
 
 
+def test_plot_metadata_secondary_haxis_scatter(ocx, api_data):
+    samples = ocx.Samples.where(project="4b53797444f846c4")
```
pkonieczny, line 91: Nit: looks like this project could be a fixture vs a hardcoded UUID?

pkonieczny (approved):
Looks great!

tests/test_viz.py
pkonieczny: Ah, the data is composed of API mocks, nevermind!

onecodex/viz/_metadata.py
jairideout: This is part of the hack to make a faceted plot look like a grouped plot: https://stackoverflow.com/a/66877669/3776794

If we don't turn off x-axis labels and ticks it looks like this (the x-axis labels are redundant with the coloring and tooltip):

Screenshot 2024-02-15 at 8 11 51 AM

jairideout: I'll add a comment to the code explaining why we're doing this

tests/test_viz.py
jairideout: Good idea, I'll make a fixture for the SampleCollection

PR #491, opened 2024-02-21
Add alpha diversity stats tests

jairideout:
## Status
- [x] **Ready**

## Description
Added stats tests for significant differences between groups of alpha diversity values:

- Wilcoxon (2 groups, paired data)
- Mann-Whitney U (2 groups, unpaired data)
- Kruskal-Wallis with optional posthoc Dunn test (>=2 groups, unpaired data)

A reasonable test is chosen automatically based on the data unless the user specified one.

The tests are run via `SampleCollection.alpha_diversity_stats()`.

Closes DEV-9236

## Related PRs
- [x] This PR is independent


onecodex/stats.py
```diff
@@ -0,0 +1,288 @@
...
+        scipy.stats.wilcoxon
+        scipy.stats.mannwhitneyu
+        scipy.stats.kruskal
+        scikit_posthocs.posthoc_dunn
+
+        """
+        if not (0.0 < alpha < 1.0):
+            raise StatsException("`alpha` must be between 0 and 1 (exclusive).")
+
+        if not isinstance(group_by, tuple):
```
audy, line 88: I think it would be good to check if `group_by` is otherwise a string. I can see myself passing a list instead of a tuple here. It might be good to also check for a list and convert it to a tuple. The same applies to `paired_by`

pkonieczny (approved):
Very elegant 🏅

onecodex/stats.py
```diff
@@ -0,0 +1,295 @@
...
+        from scipy.stats import kruskal
+
+        group_names, group_alpha_values = self._get_group_names_and_alpha_values(
+            df, metric, group_by_column_name
+        )
+        result = kruskal(*group_alpha_values)
+
+        posthoc_df = None
+        if result.pvalue < alpha and len(group_names) > 2:
+            posthoc_df = posthoc_dunn(df, val_col=metric, group_col=group_by_column_name)
```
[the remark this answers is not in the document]
lynchde: Let's use Benjamini-Hochberg please! Thanks!

onecodex/analyses.py
```diff
@@ -140,9 +150,12 @@ def _metadata_fetch(self, metadata_fields, label=None, match_taxonomy=True):
             if isinstance(f, tuple):
                 # joined categorical metadata
                 for field in f:
+                    if not isinstance(field, str):
+                        raise OneCodexException(f"Metadata field {field} must be of type str")
```
lynchde, line 154: Should we enforce strings, or could we instead stringify ints/floats?

lynchde (commented):
Just one question/possible change, but otherwise fantastic!

onecodex/analyses.py
jairideout: This is enforcing the *field name* is a string -- I'll update the error message to make that clearer!

jairideout: (string field names has always been a requirement of `_metadata_fetch()`, it would just throw a hard to interpret error previously)

PR #492, opened 2024-02-26
Add beta diversity stats tests

jairideout:
## Status
- [x] **Ready**

## Description
Added PERMANOVA test for significant differences between groups of samples based on their beta diversity distances.

Posthoc pairwise PERMANOVA tests are performed if the global test is found to be statistically significant and there are more than two groups.

The tests are run via `SampleCollection.beta_diversity_stats()`.

Samples with missing abundances are filtered out (with a warning) in:

- `plot_pcoa()`
- `plot_mds()`
- `plot_pca()`
- `alpha_diversity_stats()`
- `beta_diversity_stats()`

Deprecated `AlphaDiversityStatsResults.posthoc_df` in favor of `AlphaDiversityStatsResults.posthoc.adjusted_pvalues`.

Closes DEV-9248

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/stats.py
+++ b/onecodex/stats.py
@@ -1,13 +1,26 @@
 from __future__ import annotations
-from dataclasses import dataclass
+import itertools
 import warnings
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Optional
 
 from onecodex.exceptions import OneCodexException, StatsException, StatsWarning
-from onecodex.lib.enums import Rank, AlphaDiversityMetric, AlphaDiversityStatsTest
+from onecodex.lib.enums import (
+    Rank,
+    AlphaDiversityMetric,
+    AlphaDiversityStatsTest,
+    BetaDiversityMetric,
+    BetaDiversityStatsTest,
+)
 
 if TYPE_CHECKING:
     import pandas as pd
+    import skbio
+
+
+# TODO when min Python version is >=3.10, create a base class with
+# @dataclass(frozen=True, kw_only=True) and have AlphaDiversityStatsResults and
+# BetaDiversityStatsResults inherit from it
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


onecodex/stats.py
```diff
@@ -0,0 +1,235 @@
...
+            raise StatsException("`alpha` must be between 0 and 1 (exclusive).")
+
+        group_by = self._tuplize(group_by)
+
+        # Munge the metadata first in case there's any errors
+        df, magic_fields = self._metadata_fetch([group_by], coerce_missing_composite_fields=False)
+        group_by_column_name = magic_fields[group_by]
+
+        # Compute beta diversity distance matrix
+        dm = self.beta_diversity(metric=metric, rank=rank)
```
pkonieczny, line 104: Nit: we could probably move this test lower so that we don't run it if we know there aren't enough groups?

pkonieczny (approved):

onecodex/stats.py
```diff
@@ -255,44 +338,188 @@ def _kruskal(
...
+        Parameters
+        ----------
+        group_by : str or tuple of str or list of str
+            Metadata variable to group samples by. At least two groups are required. If `group_by`
+            is a tuple or list, field values are joined with an underscore character ("_").
+        metric : {'braycurtis', 'jaccard', 'cityblock', 'manhattan', 'aitchison', 'unweighted_unifrac', 'weighted_unifrac'}, optional
+            The beta diversity distance metric to calculate. Note that 'cityblock' and 'manhattan'
+            are equivalent metrics.
+        rank : {'auto', 'kingdom', 'phylum', 'class', 'order', 'family', 'genus', 'species'}, optional
+            Analysis will be restricted to abundances of taxa at the specified level.
```
lynchde, line 384: It looks as though, if there are any samples for which we don't have abundance estimates, we might instead be falling back to read counts. We're calculating beta diversity stats on a total sample size, rather than the sample size of just those samples with abundances.
(This is not the case for alpha diversity, where we warn that samples are excluded if they don't have abundance estimates).
I think what we should do is add a new variable to beta_diversity_stats (and also alpha_diversity_stats), similar to what we have with dataframes, with ocx_metric : {'readcount_w_children', 'readcount', 'abundance'}. (Default to abundance). Then use this for the beta diversity calculation, rather than letting beta div decide. (Open to discussion).

lynchde (changes requested):
Mostly great. Just a small change to deal with how beta diversity will default back to reads if any abundances are missing.

onecodex/stats.py
jairideout: I added filtering of samples with missing abundances to:

- `beta_diversity_stats()`
- `plot_pcoa()`/`plot_mds()`
- `plot_pca()`

Filtering was already being done for `plot_distance()`. If filtering happens, we show a warning to the user.

I'll create a followup issue to make this change more generally in `beta_diversity()` (it ended up being more complicated to add it there after all).

lynchde (approved):
Looks great!

PR #500, opened 2024-05-08
Fix calculation when all samples lack abundance estimates

audy:
## Status
- [x] **Ready**

## Description

- Fix edge-case where abundance for "Other" is shown as 1.0 when there are no samples with abundance estimates. This is due to a weird behavior in Pandas where `df.apply` yields a single empty Series when the dataframe is empty.

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Tests


alenajk (approved):
nit: Inline comment would be nice to explain why the change is needed otherwise LGTM

alenajk (approved):

PR #503, opened 2024-07-25
Add support for concatenating ONT files

pkonieczny:
## Status
- [X] **Ready**

## Description
This PR adds support for detecting and concatenating ONT file groups. The detection in based on ordinal in the file names - they need to end with `_X` (starting with 0 index) and have no gaps in the sequence.
If a user accepts (or skips the prompt altogether), such file sequences are concatenated and the resulting file is being uploaded instead of the captured sequence. 

## Related PRs
- [X] This PR is independent

### DIFF
--- a/onecodex/input_helpers.py
+++ b/onecodex/input_helpers.py
@@ -3,51 +3,140 @@ import re
 import os
 import shutil
 import tempfile
+import logging
 from collections import defaultdict
+
+# Captures parts before and after ordinal
+# (. or _ followed by num followed by . or _ and non-digits)
+ORDINAL_REV_PATTERN = r"([._])\d([._][\D._]+)$"
+
+# Captures parts before and after paired ordinal (as above, but includes R1, r1, R2, r2)
+PAIRED_ORDINAL_REV_PATTERN = r"([._][Rr])\d+([._][\w.]+)$"
+
+log = logging.getLogger("onecodex")
+
+
+def _replace_filename_ordinal(filename, replacement):
+    """Replace file[_.]num[._] with file[_.]replacement[._]."""
+    replace_pattern = rf"\g<1>{replacement}\g<2>"
+    return re.sub(ORDINAL_REV_PATTERN, replace_pattern, filename)
+
+
+def _replace_paired_filename_ordinal(filename, replacement):
+    """Replace file[_.][Rr]?num[._] with file[_.][Rr]?replacement[._]."""
+    first_pass = _replace_filename_ordinal(filename, replacement)
+    replace_pattern = rf"\g<1>{replacement}\g<2>"
+    return re.sub(PAIRED_ORDINAL_REV_PATTERN, replace_pattern, first_pass)
+
+
+def concatenate_ont_groups(files, prompt):
+    """Concatenate ONT split files and return the group as a single entry on the files list."""
+    single_files = set(files)
+    ont_groups = defaultdict(set)
+    auto_group = True
+
+    for filename in files:
+        ont_zero_filename = _replace_filename_ordinal(filename, "0")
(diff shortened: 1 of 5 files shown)


onecodex/input_helpers.py
```diff
@@ -3,51 +3,140 @@
...
+        return files
+
+    # Ensure there is no gap in the file sequences
+    for base_ont_filename, files in ont_groups.items():
+        ont_file = next(iter(files))
+        expected_sequence = [_replace_filename_ordinal(ont_file, idx) for idx in range(len(files))]
+        if any(expected_file not in files for expected_file in expected_sequence):
+            # Detected a gap in the ONT file sequence, skipping
+            log.warning(
+                f"Detected a gap in the ONT file sequence for {os.path.basename(base_ont_filename)}, skipping concatenation"
```
petraszd, line 78: Nit: It would be nice to tell an user which file(s) is(are) missing.

petraszd (approved):

onecodex/input_helpers.py
```diff
@@ -3,51 +3,140 @@
...
+    single_files = set(files)
+    ont_groups = defaultdict(set)
+    auto_group = True
+
+    for filename in files:
+        ont_zero_filename = _replace_filename_ordinal(filename, "0")
+        if os.path.exists(ont_zero_filename):
+            # strip the ordinal and preceding . or _
+            base_filename = re.sub(r"[._]\d([._][\D._]+)$", r"\1", filename)
+            base_filename = os.path.join(tempfile.gettempdir(), base_filename)
```
jairideout, line 43: It'd be nice to clean up the temporary concatenated files when the process completes

jairideout (approved):

onecodex/input_helpers.py
```diff
@@ -3,51 +3,140 @@
...
+        if any(expected_file not in files for expected_file in expected_sequence):
+            # Detected a gap in the ONT file sequence, skipping
+            log.warning(
+                f"Detected a gap in the ONT file sequence for {os.path.basename(base_ont_filename)}, skipping concatenation"
+            )
+            continue
+        with open(base_ont_filename, "wb") as outf:
+            for ont_filename in expected_sequence:
+                with open(ont_filename, "rb") as inf:
+                    shutil.copyfileobj(inf, outf)
```
audy, line 84: I think this is fine we we are assuming gzipped. If they're not gzipped; what happens if there is no newline at the end of a file?

E.g.

```
@seq1
GATC
+seq1
xxxx@seq2
GATC
+seq2
xxxx
```

onecodex/input_helpers.py
pkonieczny: Due to using `tempdir`, the file is already being removed. I've just checked and the concatenated file is not on the filesystem after the command finishes.

onecodex/input_helpers.py
pkonieczny: I think we can skip that check as I see that by default, the ONT files do have newlines at the end and they concatenate just fine

onecodex/input_helpers.py
pkonieczny: Good catch, I thought it was being cleaned up already!

alenajk (approved):

onecodex/utils.py
```diff
@@ -484,3 +487,15 @@ def is_continuous(series):
 
 def has_missing_values(dataframe_or_series):
     return dataframe_or_series.isnull().values.any()
+
+
+@contextmanager
+def use_tempdir():
```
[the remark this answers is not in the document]
pkonieczny: Good point! I still wrapped this call in `use_tempdir` so that it's easier to change up the implementation if needed

PR #512, opened 2024-11-25
Filter and warn if samples are missing abundances in `plot_metadata()`

jairideout:
## Status
- [x] **Ready**

## Description
Samples with missing abundances are filtered out in `plot_metadata()` when plotting alpha diversity or taxon abundances. A warning is also raised.

If numeric metadata is plotted on the vertical axis, this filtering is not applied because abundances are not needed to create the plot.

An error is raised if all samples are filtered out due to missing abundances.

Closes DEV-10065

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/analyses.py
+++ b/onecodex/analyses.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
 import warnings
 from collections import Counter
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 from onecodex.exceptions import OneCodexException
 from onecodex.lib.enums import (
@@ -20,6 +23,9 @@ from onecodex.viz import (
 )
 from onecodex.stats import StatsMixin
 
+if TYPE_CHECKING:
+    import pandas as pd
+
 
 def _get_all_nan_classification_ids(df):
     all_nan_classification_ids = []
@@ -27,6 +33,13 @@ def _get_all_nan_classification_ids(df):
         if is_all_nan:
             all_nan_classification_ids.append(class_id)
     return all_nan_classification_ids
+
+
+@dataclass
+class MetadataFetchResults:
+    df: pd.DataFrame  # transformed metadata
+    renamed_fields: dict  # Map of original fields to renamed fields
+    taxonomy_fields: set  # Set of *original* field names that matched taxonomy
 
 
 class AnalysisMixin(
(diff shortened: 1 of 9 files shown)


onecodex/analyses.py
```diff
@@ -29,6 +35,13 @@ def _get_all_nan_classification_ids(df):
     return all_nan_classification_ids
 
 
+@dataclass
+class MetadataFetchResults:
+    df: pd.DataFrame  # transformed metadata
+    renamed_fields: dict  # Map of original fields to renamed fields
```
petraszd, line 41: Nit: `dict[str, str]`.

petraszd (approved):

onecodex/viz/_metadata.py
```diff
@@ -151,6 +158,21 @@ def plot_metadata(
             df = pd.concat([df, vert_df], axis=1).dropna(subset=[vert_magic_fields[vaxis]])
             magic_fields.update(vert_magic_fields)
 
+        if filter_samples_missing_abundances and self._all_nan_classification_ids:
+            df = df[~df.index.isin(self._all_nan_classification_ids)]
+
+            if len(df) < 1:
```
alenajk, line 164: nit: Could this just check len(df)==0?

alenajk (approved):

onecodex/viz/_metadata.py
```diff
@@ -151,6 +158,21 @@ def plot_metadata(
             df = pd.concat([df, vert_df], axis=1).dropna(subset=[vert_magic_fields[vaxis]])
             magic_fields.update(vert_magic_fields)
 
+        if filter_samples_missing_abundances and self._all_nan_classification_ids:
```
audy, line 161:
```suggestion
        if filter_samples_missing_abundances and self._all_nan_classification_ids:
```

I think `_all_nan_classification_ids` will always be empty if you use a metric other than abundance but maybe it's clearer to explicitly also check that the metric is `Abundance`.

It might also be clearer to rename `_all_nan_classification_ids` to `_classification_ids_without_abundances`

This is mostly just me pontificating and not a big concern

audy (approved):

onecodex/analyses.py
```diff
@@ -1,5 +1,8 @@
+from __future__ import annotations
```
audy, line 1: Is this necessary for Python > 3.9?

onecodex/analyses.py
```diff
@@ -251,7 +277,9 @@ def _metadata_fetch(
                     magic_metadata[f] = None
                     magic_fields[f] = str_f
 
-        return magic_metadata, magic_fields
+        return MetadataFetchResults(
```
audy, line 280: Specify `MetadataFetchResults` as the return type

onecodex/analyses.py
jairideout: It's needed for postponed evaluation of annotations (e.g. deferred pandas imports)

PR #537, opened 2025-07-14
Have pytest fail on warnings

jairideout:
## Status
- [x] **Ready**

## Description
Also fixed all existing warnings.

Closes DEV-10683

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/dataframes.py
+++ b/onecodex/dataframes.py
@@ -122,7 +122,6 @@ class ClassificationsSeries(pd.Series):
         dtype=None,
         name=None,
         copy=False,
-        fastpath=False,
         ocx_rank=None,
         ocx_metric=None,
         ocx_taxonomy=None,
@@ -135,9 +134,7 @@ class ClassificationsSeries(pd.Series):
         self.ocx_metadata = ocx_metadata
         self.ocx_normalized = ocx_normalized
 
-        super().__init__(
-            data=data, index=index, dtype=dtype, name=name, copy=copy, fastpath=fastpath
-        )
+        super().__init__(data=data, index=index, dtype=dtype, name=name, copy=copy)
 
     @property
     def _constructor(self):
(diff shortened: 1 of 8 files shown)


onecodex/dataframes.py
```diff
@@ -135,9 +134,7 @@ def __init__(
         self.ocx_metadata = ocx_metadata
         self.ocx_normalized = ocx_normalized
 
-        super().__init__(
-            data=data, index=index, dtype=dtype, name=name, copy=copy, fastpath=fastpath
-        )
+        super().__init__(data=data, index=index, dtype=dtype, name=name, copy=copy)
```
audy, line 137: What do you think about this?

```
def __init__(self, ocx_rank=, ..., **kwargs):
   # handle ocx-specific args
   self.ocx_ = ...
   super().__init__(**kwargs)
```

For: maintain backwards compatibility, stays current with Pandas. Against: it's now possible to pass more kwargs to `pd.Series()`

onecodex/viz/__init__.py
```diff
@@ -82,38 +82,42 @@
...
-            "legend": {
-                "labelFont": font_family,
-                "labelFontSize": 12,
-                "titleFont": font_family,
-                "titleFontSize": 12,
-            },
-            "title": {"font": font_family},
-            "view": {"width": 400, "height": 400, "strokeWidth": 0},
-            "background": "white",
+    return alt.theme.ThemeConfig(
```
audy, line 91: technically breaking backwards compat. Do we need to change the return type in order to silence the warning or can `alt.theme.ThemeConfig` be called in `configure_onecodex_theme`?

petraszd (approved):

pkonieczny (approved):

onecodex/viz/__init__.py
jairideout: Altair's deprecation warning recommended using this new return type, but I just tested with returning a dict and it seems to work (and doesn't raise the warning). Reverted!

PR #616, opened 2026-04-06
Add ANCOM-BC differential abundance test

jairideout:
## Status
- [x] **Ready**

## Description
Added ANCOM-BC differential abundance test via `SampleCollection._ancombc()`. This method is currently **experimental** and breaking API changes may occur.

- Multiple metadata grouping variables are supported (e.g. secondary, tertiary, ...)
- An explicit reference group may be specified
- A global ANCOM-BC test may be run in addition to the main results (only available if there are at least 3 groups)
- Abundance metric, rank, and alpha are parametrized

Plots will be included in a future PR.

Closes DEV-11520

### Example: 2 groups

Screenshot 2026-04-14 at 2 07 41 PM

### Example: >2 groups, with secondary group by

Screenshot 2026-04-14 at 2 09 15 PM
Screenshot 2026-04-14 at 2 09 34 PM

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/stats.py
+++ b/onecodex/stats.py
@@ -108,11 +108,12 @@ class AncombcResults(StatsResults):
     """A dataclass for storing the results of an ANCOM-BC differential abundance test.
 
     - `main_results`: `pd.DataFrame` containing ANCOM-BC test results for each taxon, including
-    log2-fold change (with standard error), W-statistic, p-value, adjusted p-value, and whether or
-    not the result is statistically significant. See `skbio.stats.composition.ancombc` for details.
+    log2-fold change (with standard error), W-statistic, p-value, adjusted p-value (i.e. q-value),
+    and whether or not the result is statistically significant. See
+    `skbio.stats.composition.ancombc` for details.
     - `global_results`: `pd.DataFrame` containing global ANCOM-BC test results for each taxon,
-    including W-statistic, p-value, adjusted p-value, and whether or not the result is statistically
-    significant. See `skbio.stats.composition.ancombc` for details.
+    including W-statistic, p-value, adjusted p-value (i.e. q-value), and whether or not the result
+    is statistically significant. See `skbio.stats.composition.ancombc` for details.
     - `reference_group`: Group name used as the reference. All other groups are compared to this
     reference group.
     - `adjustment_method`: method used to adjust p-values to control the false discovery rate
@@ -704,10 +705,13 @@ class StatsMixin(DistanceMixin, BaseSampleCollection):
     ) -> AncombcResults:
         """Perform a test for differentially abundant taxa using ANCOM-BC.
 
-        Readcounts are normalized and zeros are replaced using multiplicative replacement. Results
-        indicate taxa that are significantly different in abundance across the grouping variable of
-        interest, as well as each taxon's log2-fold change (with standard error), W-statistic,
-        p-value, and adjusted p-value.
+        Zeros are replaced using multiplicative replacement if `metric` is normalized
+        (i.e. proportionalized) data, otherwise a pseudocount of 1 is added if `metric` is
+        unnormalized (i.e. count) data.
+
+        Results indicate taxa that are significantly different in abundance across the grouping
+        variable of interest, as well as each taxon's log2-fold change (with standard error),
+        W-statistic, p-value, and adjusted p-value (i.e. q-value).
 
         This method is **experimental** and breaking API changes may occur.
 
@@ -722,7 +726,7 @@ class StatsMixin(DistanceMixin, BaseSampleCollection):
             to this reference group. If not specified, the first alphabetically sorted group name
             will be used as the reference.
 
-        metric: :class:`~onecodex.lib.enums.Metric`, optional
+        metric : :class:`~onecodex.lib.enums.Metric`, optional
             The taxonomic abundance metric to use. See :class:`~onecodex.lib.enums.Metric`
             for definitions.
 
@@ -815,10 +819,14 @@ class StatsMixin(DistanceMixin, BaseSampleCollection):
         # Filter the taxa dataframe to match the remaining IDs in the metadata dataframe
         taxa_df = taxa_df.loc[metadata_df.index]
 
-        # Need to use pseudocounts or `multi_replace()` because ANCOM-BC can't handle zeros
-        multi_replace_df = pd.DataFrame(
-            multi_replace(taxa_df), columns=taxa_df.columns, index=taxa_df.index
-        )
+        # ANCOM-BC can't handle zeros. Use multiplicative replacement for normalized
+        # (i.e. proportionalized) data, or a pseudocount for unnormalized (i.e. count) data.
+        if metric.is_normalized:
+            zero_replaced_df = pd.DataFrame(
+                multi_replace(taxa_df), columns=taxa_df.columns, index=taxa_df.index
+            )
+        else:
+            zero_replaced_df = taxa_df + 1
 
         # Use a safe internal column name for the patsy formula so that it works as both a
         # valid formula term and a metadata column lookup (skbio's `grouping` parameter is
@@ -828,8 +836,9 @@ class StatsMixin(DistanceMixin, BaseSampleCollection):
         while ancombc_col in metadata_df.columns:
             ancombc_col = f"_{ancombc_col}_"
         ancombc_metadata = metadata_df.rename(columns={group_by_column_name: ancombc_col})
+
         ancombc_results = ancombc(
-            multi_replace_df,
+            zero_replaced_df,
             ancombc_metadata,
             formula=ancombc_col,
             grouping=ancombc_col if include_global_test else None,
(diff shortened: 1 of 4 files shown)


onecodex/stats.py
```diff
@@ -655,3 +690,220 @@ def _permanova(
...
+            global_results=global_results,
+            reference_group=reference_group,
+            alpha=alpha,
+            adjustment_method=AdjustmentMethod.BenjaminiHochberg,
+            sample_size=len(metadata_df),
+            group_by_variable=group_by_column_name,
+            group_sizes=self._get_group_sizes(metadata_df, group_by_column_name),
+        )
+
+    def _rename_tax_ids(self, taxa_df: pd.DataFrame) -> pd.DataFrame:
```
audy, line 883: This function has been on my wishlist for a while :)

onecodex/stats.py
```diff
@@ -655,3 +690,220 @@ def _permanova(
...
+        return taxa_df.rename(
+            columns=lambda tax_id: f"{self.taxonomy['name'][tax_id]} ({tax_id})"
+            if tax_id in self.taxonomy["name"]
+            else tax_id
+        )
+
+    def _drop_samples_with_zero_abundance(
```
audy: "Abundance" has a special meaning and there's already code for dealign with "samples w/o abundance" so I would rename this. Maybe `_drop_samples_with_all_zeros`?

onecodex/stats.py
```diff
@@ -655,3 +690,220 @@ def _permanova(
...
+        self, metadata_df: pd.DataFrame, taxa_df: pd.DataFrame
+    ) -> pd.DataFrame:
+        prev_count = len(metadata_df)
+        metadata_df = metadata_df.loc[(taxa_df != 0).any(axis=1)]
+        num_dropped = prev_count - len(metadata_df)
+        if num_dropped > 0:
+            warnings.warn(
+                f"{num_dropped} sample{'s were' if num_dropped > 1 else ' was'} excluded from the "
+                f"test as {'they' if num_dropped > 1 else 'it'} had zero abundance across all "
```
audy: same here. Maybe `had zero {self.metric} values across all taxa`

onecodex/stats.py
```diff
@@ -655,3 +690,220 @@ def _permanova(
...
+        )
+
+        # ANCOM-BC can only run a global test if there's at least 3 groups
+        if include_global_test:
+            self._assert_min_num_groups_for_global_test(metadata_df, group_by_column_name)
+
+        # Filter the taxa dataframe to match the remaining IDs in the metadata dataframe
+        taxa_df = taxa_df.loc[metadata_df.index]
+
+        # Need to use pseudocounts or `multi_replace()` because ANCOM-BC can't handle zeros
```
audy: When is it appropriate to use pseudocounts?

I noticed that this scales everything to (0, 1). For normalized readcounts and abundances, this doesn't change the non-zero values much. But for readcounts, it does (I think it equivalent to our normalized readcounts). Does that mean we need to use pseudocounts for `readcount{_w_children}`?

onecodex/stats.py
```diff
@@ -655,3 +690,220 @@ def _permanova(
...
+        multi_replace_df = pd.DataFrame(
+            multi_replace(taxa_df), columns=taxa_df.columns, index=taxa_df.index
+        )
+
+        # Use a safe internal column name for the patsy formula so that it works as both a
+        # valid formula term and a metadata column lookup (skbio's `grouping` parameter is
+        # used for both). Using patsy's quoting, e.g. `Q('foo bar')`, doesn't work with the
+        # `grouping` parameter.
+        ancombc_col = "__ancombc_group__"
```
audy: is it overly paranoid to check make sure that isn't already a metadata column 😆 ?

audy (approved):
Looks good! I left some comments. My only (minor) concern is renaming `_drop_samples_with_zero_abundance`

pkonieczny (approved):
A bit out of my depth here, but this looks good to me!

onecodex/stats.py
```diff
@@ -655,3 +690,220 @@ def _permanova(
...
+        ancombc_metadata = metadata_df.rename(columns={group_by_column_name: ancombc_col})
+        ancombc_results = ancombc(
+            multi_replace_df,
+            ancombc_metadata,
+            formula=ancombc_col,
+            grouping=ancombc_col if include_global_test else None,
+            alpha=alpha,
+            p_adjust="fdr_bh",
```
jairideout: @lynchde scikit-bio defaults to Holm-Boniferroni. I'm overriding that default here to use Benjamini-Hochberg because we're using that adjustment method in alpha and beta diversity post-hoc tests. Does that make sense to you?

onecodex/stats.py
jairideout: fixed

onecodex/stats.py
jairideout: Great question. Here's what I could find:

- `multi_replace()` (re)normalizes the input data to `(0, 1)`, even if you give it counts or proportions that don't sum to 1
- The original ANCOM-BC R implementation only accepts raw counts as input, and adds a pseudocount (default of 1) to every value. Then, internally, it normalizes the counts data during the bias estimation step
- scikit-bio's implementation accepts raw counts or proportions, as long as there's no zeros
- From researching, I can't find a definitive answer about when to use one over the other

What if we add a pseudocount (default 1, parameterized) if the user specifies a raw count metric, and otherwise we use `multi_replace()` if the metric is proportional? That way we'll match the original R implementation when dealing with raw counts, and there's *some* difference between passing `metric="readcount"` vs `metric="normalized_readcount"`.

Note: I'm out of my depth here. @lynchde WDYT?

onecodex/stats.py
lynchde: There is a little bit of a trade-off between Holm-Bonferroni and Benjamini-Hochberg:
- Holm-Bonferroni is a more strict approach, and reduces false positives.
- Benjamini-Hochberg: allows for more discovery.

Given the arguments put forth by the ancom-bc creators, I'm inclined to say we should use Holm-Bonferroni for ancom-bc. (I'm ok with using BH elsewhere).

onecodex/stats.py
```diff
@@ -655,3 +690,222 @@ def _permanova(
...
+
+    def _rename_ancombc_covariate(
+        self,
+        covariate: str,
+        ancombc_col: str,
+        group_by_column_name: str,
+        reference_group: str,
+    ) -> str:
+        # Format is "__ancombc_variable__[T.<group_name>]". Reformat to
+        # "<group_by_column_name>: <group_name> vs <reference_group> (reference)"
```
jairideout, line 941: @lynchde as a general user of the client lib (e.g. in Notebooks), which covariate format do you prefer in the main results dataframe?

1. scikit-bio's `ancombc()` covariate format, e.g. `Site[T.Gut]`, `Site[T.Skin]`
2. Our own format, e.g. `Site: Gut vs Oral (reference)`, `Site: Skin vs Oral (reference)`

Option 1 is more compact and easier to manipulate via code (and matches scikit-bio). Option 2 is more human-readable and includes the reference group.

Note: for the bargraph plot and in Custom Plots, we can use option 2. I'm just not sure which format makes more sense for the main results dataframe we return here.

Another option is leaving `Covariate` in the scikit-bio format, and adding a new column (e.g. `Comparison`) that has the human-readable format.

WDYT?

onecodex/stats.py
lynchde: Great catch, @audy . And great suggestion @jairideout .
Since we don't want the readcounts to be normalized in this instance, if we can use pseudocounts for `readcount{_w_children}`, and use `multi_replace` for our other proportional metrics (`abundance{_w_children}` and `normalized_readcount{_w_children}`), then this would solve the issue.
(Background, the readcounts will be normalized differently (0-centered) within the test itself, so we shouldn't pre-normalize them if the user chooses raw readcounts).

onecodex/stats.py
lynchde: I can see benefits to both, so I think we should go with your suggestion of leaving the `Covariate` in the scikit-bio format, and adding a new column with the human-readable format. Thanks!

onecodex/stats.py
```diff
@@ -655,3 +690,222 @@ def _permanova(
...
+        reference_group: str | None = None,
+        metric: Metric = Metric.Auto,
+        rank: Rank = Rank.Auto,
+        alpha: float = 0.05,
+        include_global_test: bool = False,
+        require_classification_version_match: bool = True,
+    ) -> AncombcResults:
+        """Perform a test for differentially abundant taxa using ANCOM-BC.
+
+        Readcounts are normalized and zeros are replaced using multiplicative replacement. Results
```
lynchde: Given our decision to treat readcounts differently from proportional metrics, we'll need to update this text.

onecodex/stats.py
```diff
@@ -655,3 +690,222 @@ def _permanova(
...
+        indicate taxa that are significantly different in abundance across the grouping variable of
+        interest, as well as each taxon's log2-fold change (with standard error), W-statistic,
+        p-value, and adjusted p-value.
```
lynchde: Small suggestion to explain "adjusted p-value (i.e. q-value)"

onecodex/stats.py
```diff
@@ -655,3 +690,222 @@ def _permanova(
...
+
+        This method is **experimental** and breaking API changes may occur.
+
+        Parameters
+        ----------
+        group_by : str or tuple of str or list of str
+            Metadata variable to group samples by. At least two groups are required. If `group_by`
+            is a tuple or list, field values are joined with an underscore character ("_").
```
lynchde, line 722: Since ancombc can take multiple fields,  in the future, we might want to expand functionality (to allow for more complex models). At that point, we'll change how this works.

onecodex/stats.py
audy: > in the original ANCOM-BC R, ... it normalizes the counts data during the bias estimation step

So if we use pseudocounts + the scikit-bio version, there will be no normalization for differences in sequencing depth? If so, I think we should just keep using `multi_replace` OR only allow running stats using normalized metrics ( where `Metric.is_normalized` is `True`).

lynchde (dismissed):
This is looking great. We discussed making a few changes, mainly around how to deal with read count data, and adding a "Comparison" column. I've captured a few small things in comments. But otherwise it's looking great.

onecodex/stats.py
lynchde: From what I can tell, part of the ancombc test involves normalizing around zero. The original ancom-bc (R) implementation uses centered log ratio (CLR) normalization, which accounts for the read count differences, and is a preferred normalization method compared to straight proportions, since proportions mean "when one thing goes up, everything else must go down". This is why the original R implementation of ancom-bc only works for read counts, and why read counts are preferred as an input over abundances/normalized read counts. So I don't think we need to worry about the read counts, and the +1 method should work for those.

Newer implementations can take proportional data, and since our abundance estimates are genome-size adjusted, I still think it's valid to use our abundance/normalized data as an input (after `multi_replace`).

PR #617, opened 2026-04-07
Add ANCOM-BC results plots

jairideout:
## Status
- [x] **Ready**

## Description
Added `AncombcResults.plot()` method, which generates a bargraph of log2-fold change for each significantly different taxon, faceted by group.

Closes DEV-11520

Screenshot 2026-04-14 at 2 36 30 PM

## Related PRs
- [x] Please see the following related PRs: https://github.com/onecodex/onecodex/pull/616

### DIFF
--- a/onecodex/stats.py
+++ b/onecodex/stats.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Optional
 
 from onecodex.distance import DistanceMixin
-from onecodex.exceptions import OneCodexException, StatsException, StatsWarning
+from onecodex.exceptions import OneCodexException, PlottingException, StatsException, StatsWarning
 from onecodex.lib.enums import (
     Metric,
     Rank,
@@ -18,8 +18,10 @@ from onecodex.lib.enums import (
     PosthocStatsTest,
 )
 from onecodex.models.base_sample_collection import BaseSampleCollection
+from onecodex.viz._primitives import prepare_props
 
 if TYPE_CHECKING:
+    import altair as alt
     import pandas as pd
     import skbio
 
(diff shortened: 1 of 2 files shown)


onecodex/stats.py
```diff
@@ -127,6 +129,109 @@ class AncombcResults(StatsResults):
...
+            tooltip=["Taxon", "Covariate", "Log2(FC)", "pvalue", "qvalue"],
+        )
+
+        # Invisible points at ±max(|Log2(FC)|, 0.5) to force x-axis symmetric around 0
+        # with a minimum extent of ±0.5. Log2(FC) of 0.5 is sometimes considered
+        # "biologically relevant", which is why we use this minimum threshold.
+        sym_pos = (
+            base.mark_point(opacity=0)
+            .transform_calculate(val="max(abs(datum['Log2(FC)']), 0.5)")
+            .encode(x="val:Q", y="Taxon:O")
```
pkonieczny: I'm not sure I follow why `x` and `y` are named this way, do we always have taxons on y axis?

onecodex/stats.py
```diff
@@ -127,6 +129,109 @@ class AncombcResults(StatsResults):
...
+
+        chart = (
+            alt.layer(bars, sym_pos, sym_neg, rule)
+            .properties(**inner_props)
+            .facet(
+                row=alt.Row(
+                    "Covariate:N", title=None, header=alt.Header(labelOrient="top", labelAngle=0)
+                )
+            )
+            .resolve_scale(x="independent", y="independent")
```
pkonieczny, line 218: I think we only set `x="independent"` so far, why are we setting both here?

pkonieczny (approved):
Some minor questions, but it looks good!

onecodex/stats.py
jairideout: For these plots, yes we always have taxa on the y-axis. I'm not sure I followed your question though, let me know if that doesn't answer it

onecodex/stats.py
jairideout: Each facet (subplot) can have a different set of taxa (e.g. subplot A might have 5 significant taxa while subplot B has 12 completely different ones). Without `y="independent"`, Altair will use a shared y-axis scale across all facets, meaning every facet would show the union of all taxa from all subplots, with most bars being empty in each facet.

onecodex/stats.py
pkonieczny: Makes sense!

onecodex/stats.py
```diff
@@ -128,6 +130,98 @@ class AncombcResults(StatsResults):
...
+            )
+
+        df = df.reset_index()
+
+        df["Difference from reference"] = "No change"
+        df.loc[(df["Log2(FC)"] < 0), "Difference from reference"] = "Decreased"
+        df.loc[(df["Log2(FC)"] > 0), "Difference from reference"] = "Increased"
+
+        color_domain = ["Decreased", "Increased"]
+        color_range = ["#CC3300", "#0033CC"]
```
petraszd, line 183: Super Nit: maybe use colors from `DEFAULT_PALETTES `?

Image

petraszd (approved):

onecodex/stats.py
jairideout: @lynchde LMK if you'd like the colors updated (I used the colors from your example notebook)

lynchde: Ahh cool. I just picked a red and blue randomly. Didn't think about a specific palette, sorry. I'm happy to choose a red and blue from that if we'd prefer. Whatever is easiest. Ones from the default palette that we could choose:
- #DD3A3A for the red
- #16347B for the blue

jairideout: Updated here: https://github.com/onecodex/onecodex/pull/624

PR #621, opened 2026-04-16
Add unfiltered readcount metrics to CPITB

audy:
## Status
- [x] **Ready**
- [x] **Hold** -- need to deploy v1 API changes first

## Description

- Add two 4 new metrics for raw (unfiltered) readcounts
- Switch to the `/raw_results` endpoint because this includes the new raw readcount data (in addition to the data that's currently being retrieved)
- Warn if fetching data / plotting mixed-abundance samples and a filtered readcount metric as readcounts may not be comparable in this case
- Add the new metrics to custom plots

## Related PRs

- [x] Please see the following related PRs:

This requires changes to the v1 API.

## TODOs

- [x] Tests
- [x] Documentation
- [ ] Deploy v1 API changes


onecodex/models/analysis.py
```diff
@@ -45,10 +61,8 @@ def results(self, json=True):
         else:
             raise NotImplementedError("No non-JSON result format implemented.")
 
+    @lru_cache
     def _results(self):
-        if getattr(self, "_cached_result", None):
-            return self._cached_result
-
         resp = self._client.get(f"{self._api._base_url}{self.field_uri}/results")
         self._cached_result = resp.json()
```
jairideout: Do we still need `self._cached_result` now that you're using `@lru_cache`?

onecodex/models/analysis.py
```diff
@@ -198,7 +212,16 @@ class Classifications(_AnalysesBase, ClassificationSchema):
...
     _NONSPECIFIC_TAX_IDS = {"1", "131567"}
 
-    def results(self, json=True):
+    @lru_cache
+    def _results(self, raw: bool = False):
+        if raw:
+            resp = self._client.get(f"{self._api._base_url}{self.field_uri}/raw_results")
+        else:
+            resp = self._client.get(f"{self._api._base_url}{self.field_uri}/results")
+        self._cached_result = resp.json()
```
jairideout: same question here

onecodex/models/base_sample_collection.py
```diff
@@ -415,6 +415,35 @@ def metadata(self) -> pd.DataFrame:
...
+
+        Returns a DataFrame indexed by tax_id with columns: name, rank, parent_tax_id.
+        """
+        import pandas as pd
+
+        tax_info: dict[str, list] = {"tax_id": [], "name": [], "rank": [], "parent_tax_id": []}
+        tax_ids: set[str] = set()
+
+        for classification in self._classifications:
+            results = classification.results(raw=True)
```
jairideout, line 429: Do we always want `raw=True` here?

onecodex/models/base_sample_collection.py
```diff
@@ -442,32 +471,24 @@ def _collate_results(
...
+        # and the samples contain a mix of +/- abundance samples, then the metric may not be
+        # comparable. warn the user and advise to use the unfiltered readcounts instead
+        if metric.is_filtered_readcount_metric and (
+            0 < len(self._classification_ids_without_abundances) < len(self._classifications)
+        ):
+            warnings.warn(
+                f"{len(self._classification_ids_without_abundances)} sample(s) have no abundances "
+                f"calculated. {metric.display_name} values may not be comparable across samples when abundance "
+                "status is mixed. Consider using an unfiltered metric instead.",
+                PlottingWarning,
```
jairideout: Can you raise a different warning class? `PlottingWarning` doesn't seem appropriate here because this functionality isn't specific to plots

onecodex/viz/_custom_plots/utils.py
```diff
@@ -82,12 +82,16 @@ def format_classification_results(results: dict) -> dict[str, Any]:
...
             "abundance_w_children": data.get("c") if has_abundances else None,
+            "raw_readcount": data.get("x"),
+            "raw_readcount_w_children": data.get("y"),
         }
 
         # Skip 0 values (i.e., reads with only k-mer hits but no reads or child reads)
         if (
             entry["readcount"] < 1
             and entry["readcount_w_children"] < 1
+            and entry["raw_readcount"] < 1
```
jairideout: I think this could raise a `TypeError` if comparing `None < 1` (same comment for the line below)

tests/test_collection.py
```diff
@@ -125,6 +125,22 @@ def test_sample_collection_pandas(samples):
             Rank.Species,
             [0.1349, 0.0231, 0.0087, 0.0069, 0.0065, 0.0065],
         ),
+        # NormalizedRaw mirrors Normalized since mock data has raw_readcount == readcount
```
jairideout: nit: maybe we could make the raw readcounts different, so we can detect bugs where we're accessing filtered vs unfiltered results incorrectly

tests/test_viz.py
```diff
@@ -389,8 +389,12 @@ def test_plot_pca_missing_abundances(ocx, api_data, samples, samples_without_abu
     with pytest.warns(PlottingWarning, match=r"3 sample\(s\) have no abundances calculated"):
         samples.plot_pca(metric="abundance", return_chart=True)
 
-    # non abundance-based metric, no warning
-    samples.plot_pca(metric="readcount", return_chart=True)
+    # filtered readcount metric warns about mixed abundance status
+    with pytest.warns(PlottingWarning, match=r"3 sample\(s\) have no abundances calculated"):
```
jairideout: nit: make the `match` regex more specific so we can differentiate the new "mixed abundance status" message from the check on line 389

onecodex/lib/enums.py
```diff
@@ -109,6 +109,34 @@ class Metric(BaseEnum):
        .. math::
 
           \frac{\text{readcount\_w\_children}}{\sum_{\text{taxa at rank}} \text{readcount\_w\_children}}
+
+    .. attribute:: RawReadcount
```
jairideout: naming nit: maybe it's clearer to call these `UnfilteredReadcount`, etc. Feel free to ignore

audy: I actually like that better 👍

onecodex/models/analysis.py
audy: 👍 I'm going to go ahead and replace all of our custom cache instances with `@lru_cache`

onecodex/models/base_sample_collection.py
audy: Yes because we cache the fetch from the API. If we mix `raw=True/False` then we could end up making unnecessary API calls.

onecodex/models/base_sample_collection.py
audy: added `OneCodexWarning`. I will update CPITB to catch and display this as well.

onecodex/models/analysis.py
```diff
@@ -212,22 +226,20 @@ def results(self, json=True):
...
         return resp.json()
 
-    def _table(self):
+    def _table(self, raw: bool = False) -> pd.DataFrame:
         import pandas as pd
 
-        if self._cached_table is None:
-            self._cached_table = pd.DataFrame(self._results()["table"])
-        return self._cached_table
+        return pd.DataFrame(self._results(raw=raw)["table"])
```
audy, line 240: `pd.DataFrame(x)` is probably fast enough that we don't need to cache it.

onecodex/viz/_custom_plots/collection.py
```diff
@@ -243,6 +250,7 @@ def _run_with_plot_error_handling(self, plot_fn: Callable[[], PlotResults]) -> P
 
         with warnings.catch_warnings(record=True) as captured_warnings:
             warnings.simplefilter("always", PlottingWarning)
+            warnings.simplefilter("always", OneCodexWarning)
```
audy: alternatively, we could capture `UserWarning` (which also includes `StatsWarning`). WDYT @jairideout ?

audy: Changed to:

```
UserWarning > OneCodexUserWarning > {StatsWarning,PlottingWarning}
```

jairideout (approved):

PR #623, opened 2026-04-17
Disallow numeric group by variable in ANCOM-BC

jairideout:
## Status
- [x] **Ready**

## Description
`_ancombc()` now raises a `StatsException` if the selected group by variable is numeric.

scikit-bio's `ancombc()` converts the variable to `float64` (even if passed as a string), and uses linear regression instead of categorical pairwise group comparisons. The output format is also a bit different when using a numeric variable. For now, restrict the test to only allow categorical variables.

In the future, we might want to allow users to specify whether the variable is categorical or numeric. I explored some hacks to get that to work, and it's *possible* but pretty complicated.

Closes DEV-11581

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/stats.py
+++ b/onecodex/stats.py
@@ -891,6 +891,8 @@ class StatsMixin(DistanceMixin, BaseSampleCollection):
         if require_classification_version_match:
             self._assert_single_database_version(metadata_df)
 
+        self._assert_categorical_group_by(metadata_df, group_by_column_name)
+
         group_names = sorted(metadata_df[group_by_column_name].unique())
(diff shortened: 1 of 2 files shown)


audy (approved):

petraszd (approved):

pkonieczny (approved):

lynchde (approved):
This is good for MVP. Eventually we will want to allow numerical variables, but we will also want to confirm if the customer intends for the number to be continuous or categorical, which will be another situation. For now, this forces that decision.

PR #630, opened 2026-04-28
Bugfix: plot_bargraph errors out if result df is empty

petraszd:
## Status
- [x] **Ready**

## Description

If `df` is "almost empty" then `df.apply` may be still called (for some pandas internal reasons) even if there are zero rows and `row_totals` are empty iterator. Thus causing `StopIterator` error.

## Related PRs
- [x] This PR is independent

### DIFF
--- a/onecodex/viz/_bargraph.py
+++ b/onecodex/viz/_bargraph.py
@@ -170,8 +170,9 @@ class VizBargraphMixin(BaseSampleCollection):
             df = df.loc[:, df.mean().sort_values(ascending=False).iloc[:top_n].index]
 
         if include_other:
-            # It needs to be a list and not an index
-            df_index = list(df.index)
+            row_totals_by_name = {}
+            for (_, row), row_total in zip(df.iterrows(), row_totals):
+                row_totals_by_name[row.name] = row_total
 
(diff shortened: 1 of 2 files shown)


onecodex/viz/_bargraph.py
```diff
@@ -171,14 +170,19 @@ def plot_bargraph(
...
-                total = next(row_totals)
                 # if there are no abundances in the dataframe, df.apply will yield
                 # a single item that has `None` as its name. Therefore, we must
                 # check if row.name is None AND whether or not the row name is
                 # in empty_rows
-                return 0.0 if (row.name is None or row.name in empty_rows) else total - row.sum()
+                if row.name is None or row.name in empty_rows:
+                    return 0.0
+
+                total = row_totals[df_index.index(row.name)]
```
jairideout: `df_index.index()` is O(N), can you change it to use a `dict` as lookup?

alenajk (approved):

onecodex/viz/_bargraph.py
petraszd: Done.

jairideout (approved):

PR #638, opened 2026-05-14
Fetch results directly from static file with API fallback

audy:
## Status

- [x] **Ready**

## Description

Note: **adds zstandard to dependencies**

Fetches processed results from `results_uri`, bypassing the v1 API. If `results_uri` is `None` (which is currently true for incomplete analyses); falls back to the API `/results` endpoint. This is to preserve backwards compatibility for users who are calling `.results()` in a "polling" pattern:

```python
while True:
    try:
        results = run.results()
    except:
        time.sleep(10)
    else:
        break
```

Otherwise, we would need to refresh the classification instance (hitting the v1 API) every time `.results()` was called. That could maybe be reduced to only when the signed URI expires but that's more complex and we can always add that as a future optimization.

I also added support for non-compressed and gzipped files in case we ever want to migrate away from zstandard in the future (in case there are complications with pyodide).

Benchmark results (fetching 10 classification runs). There's a ~2x speedup for the client.

```
  ┌────────────────────────┬───────────────────┬────────┬─────────────┐
  │        Scenario        │ Steady-state mean │  Std   │ vs baseline │
  ├────────────────────────┼───────────────────┼────────┼─────────────┤
  │ Baseline (master)      │ 0.83s             │ ±0.03s │ —           │
  ├────────────────────────┼───────────────────┼────────┼─────────────┤
  │ Just zstd + S3 session │ 0.48s             │ ±0.01s │ -42%        │
  ├────────────────────────┼───────────────────┼────────┼─────────────┤
  │ Just job name fix      │ 0.82s             │ ±0.17s │ -1%         │
  ├────────────────────────┼───────────────────┼────────┼─────────────┤
  │ Both (PR)              │ 0.47s             │ ±0.02s │ -43%        │
  └────────────────────────┴───────────────────┴────────┴─────────────┘
```

### Also in this PR

- Fix: redundant calls to `/jobs` API being made to fetch `job.name` per classification.

## Related PRs

- [x] Please see the following related PRs: the API

## TODOs

- [x] Tests
- [x] Documentation


pkonieczny (approved):
Nice!

onecodex/models/analysis.py
```diff
@@ -207,6 +220,26 @@ class Classifications(_AnalysesBase, ClassificationSchema):
...
+        # results_uri is a pre-signed URL included in the API response; it's None when the
+        # analysis isn't available. Fetch directly from the URL when present to avoid the round-trip
+        # through the API server. Local paths are supported for testing.
+        if self.results_uri is not None:
+            import orjson
+
+            uri = self.results_uri
+            if uri.startswith(("http://", "https://")):
+                try:
+                    resp = requests.get(uri)
```
jairideout: It might make sense to use `onecodex.utils.get_requests_session()` for the retries. We might want to also add a timeout so it doesn't wait forever

jairideout (approved):

PR #639, opened 2026-05-18
Add ability to run workflows

pkonieczny:
## Status
- [X] **Ready**

## Description
This PR adds ability to run Workflows using `Jobs.run` method which returns an `Analyses` object representing the new run. This PR also allows for refreshing `Analyses` object state using a `refresh` method.
Also in this PR: adding the run Workflows option and Workflow status to the CLI.

## Related PRs
- [X] This PR is independent

### DIFF
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -699,7 +699,7 @@ def jobs_run(ctx, job_id, sample_id, args, dependency_overrides, populate_defaul
                 f"Expected <analysis_id> or <analysis_id>=<download_path>, got {dep!r}.",
                 param_hint="-d/--dependency-override",
             )
-        dep_analysis = job = ctx.obj["API"].Analyses.get(analysis_id)
+        dep_analysis = ctx.obj["API"].Analyses.get(analysis_id)
         parsed_dependencies.append(
             DependencyOverride(analysis=dep_analysis, download_path=download_path or None)
         )
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -636,12 +636,12 @@ def logout(ctx):
 
 
 # job run
-@onecodex.group("workflow", help="One Codex platform Workflow management.")
-def workflow_group():
+@onecodex.group("jobs", help="One Codex platform Workflow management.")
+def jobs_group():
     pass
 
 
-@workflow_group.command("run")
+@jobs_group.command("run")
 @click.argument(
     "job_id",
     nargs=1,
@@ -678,7 +678,7 @@ def workflow_group():
 @pretty_errors
 @telemetry
 @login_required
-def workflow_run(ctx, job_id, sample_id, args, dependency_overrides, populate_default_arguments):
+def jobs_run(ctx, job_id, sample_id, args, dependency_overrides, populate_default_arguments):
     """Run a OneCodex workflow with optional arguments."""
     from onecodex.models.misc import DependencyOverride
 
@@ -712,47 +712,5 @@ def workflow_run(ctx, job_id, sample_id, args, dependency_overrides, populate_de
         dependency_overrides=parsed_dependencies or None,
         populate_default_arguments=populate_default_arguments,
     )
-    click.echo(f"Job run successfully. New run ID: {run.id}")
-    click.echo(f"Get its status using `onecodex workflow status {run.id}`")
-
-
-@workflow_group.command("status")
-@click.argument(
-    "run_id",
-    nargs=1,
-    required=True,
-)
-@click.pass_context
-@pretty_errors
-@login_required
-def workflow_status(ctx, run_id):
-    """Get the status of an OneCodex workflow."""
-    run = ctx.obj["API"].Analyses.get(run_id)
-
-    if not run.complete:
-        status = "Running"
-    elif run.success:
-        status = "Succeeded"
-    else:
-        status = "Failed"
-
-    fields = [
-        ("Run ID", run.id),
-        ("Status", status),
-        ("Analysis type", run.analysis_type),
-        ("Job", run.job.id),
-        ("Sample", run.sample.id),
-        ("Created", run.created_at.isoformat()),
-    ]
-    if run.error_msg:
-        fields.append(("Error", run.error_msg))
-
-    width = max(len(label) for label, _ in fields)
-    for label, value in fields:
-        click.echo(f"{label + ':':<{width + 2}}{value}")
-
-    if run.job_args:
-        click.echo(f"{'Args:':<{width + 2}}", nl=False)
-        pad = " " * (width + 2)
-        for i, (key, value) in enumerate(run.job_args.items()):
-            click.echo(f"{'' if i == 0 else pad}{key}={value}")
+    click.echo(f"Job run successfully. New workflow ID: {run.id}")
+    click.echo(f"Get its status using `onecodex workflows {run.id}`")
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -633,3 +633,92 @@ def login(ctx):
 def logout(ctx):
     """Delete an API key (saved in ~/.onecodex)."""
     _logout()
+
+
+# job run
+@onecodex.group("workflow", help="One Codex platform Workflow management.")
+def workflow_group():
+    pass
+
+
+@workflow_group.command("run")
+@click.argument(
+    "job_id",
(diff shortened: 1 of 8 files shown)


onecodex/models/analysis.py
```diff
@@ -24,27 +25,24 @@
...
-        import gzip
-
-        return gzip.decompress(data)
-    return data
-
-
 class _AnalysesBase(OneCodexBase):
     _allowed_methods = {
         "instances_public": None,
+        "update": AnalysisUpdateSchema,
```
alenajk: Is there another way we could get the schema to refresh Analysis objects that's clearer/separate from which methods are user-callable?

jairideout: Why are we setting `update` if updates aren't allowed?

onecodex/models/misc.py
```diff
@@ -18,6 +19,16 @@
...
 )
 
+if TYPE_CHECKING:
+    from onecodex.models.sample import Samples
+    from onecodex.models.analysis import Analyses
+
+
+@dataclass(frozen=True)
+class DependencyOverride:
+    analysis_id: str
```
jairideout: Maybe this should be an `Analyses` object instead? Or support both `Analyses` and ID?

onecodex/cli.py
```diff
@@ -633,3 +633,84 @@ def login(ctx):
...
+@login_required
+def jobs_run(ctx, job_id, sample_id, args, dependency_overrides, populate_default_arguments):
+    """Run a OneCodex workflow with optional arguments."""
+    from onecodex.models.misc import DependencyOverride
+
+    parsed_args = {}
+    for a in args:
+        if "=" not in a:
+            raise click.BadParameter(
+                "Expected key=value format, got {!r}.".format(a), param_hint="-a/--arg"
```
jairideout: nit: use f-strings

onecodex/cli.py
```diff
@@ -633,3 +633,84 @@ def login(ctx):
...
+    job = ctx.obj["API"].Jobs.get(job_id)
+    sample = ctx.obj["API"].Samples.get(sample_id)
+
+    run = job.run(
+        sample,
+        parsed_args,
+        dependency_overrides=parsed_dependencies or None,
+        populate_default_arguments=populate_default_arguments,
+    )
+    click.echo(f"Job run successfully. New workflow ID: {run.id}")
```
jairideout: "Job run created successfully. New analysis ID: {run.id}"

onecodex/cli.py
```diff
@@ -633,3 +633,84 @@ def login(ctx):
...
+    click.echo(f"Get its status using `onecodex workflows {run.id}`")
```
jairideout: "Get its status using `onecodex analyses {run.id}`"

(not all jobs will produce WorkflowRuns)

onecodex/cli.py
```diff
@@ -633,3 +633,84 @@ def login(ctx):
 def logout(ctx):
     """Delete an API key (saved in ~/.onecodex)."""
     _logout()
+
+
+# job run
+@onecodex.group("jobs", help="One Codex platform Workflow management.")
```
jairideout: "One Codex platform Job management."

onecodex/cli.py
```diff
@@ -633,3 +633,84 @@ def login(ctx):
...
+    "--populate-default-arguments/--no-populate-default-arguments",
+    default=True,
+    help="Populate job arguments with their defaults on the server (default: enabled).",
+)
+@click.pass_context
+@pretty_errors
+@telemetry
+@login_required
+def jobs_run(ctx, job_id, sample_id, args, dependency_overrides, populate_default_arguments):
+    """Run a OneCodex workflow with optional arguments."""
```
jairideout: """Run a OneCodex job with optional arguments."""

onecodex/cli.py
```diff
@@ -633,3 +633,84 @@ def login(ctx):
...
+    from onecodex.models.misc import DependencyOverride
+
+    parsed_args = {}
+    for a in args:
```
alenajk: nit: I'd prefer `arg` over `a` as a var name

onecodex/cli.py
alenajk: nit: Maybe update the wording to make it clear that it was kicked off/started successfully ("run successfully" could make it sound like it finished?)

alenajk (approved):

onecodex/models/analysis.py
pkonieczny: The internals of our models prevent `setattr` if `update` is not set. It can be worked around using `self.__dict__[attr] = X` but then it's just hacking the Python language 🤷 

We could also just _not_ have a `refresh` method and instead rely on fetching the Analysis object anew each time - but that feels like bad design for the user.

pkonieczny: I'm open to other ideas though!

onecodex/models/misc.py
pkonieczny: Good point, updated!

onecodex/models/analysis.py
pkonieczny: I don't think I follow?

alenajk: Just wondering if there's another way to accomplish the goal besides adding it to `_allowed_methods` when it's not actually allowed

jairideout: Ah I see, the comment was confusing me a little. Do you mind updating the comment to say that we need to allow updates in order to use `setattr`?

onecodex/models/misc.py
```diff
@@ -26,7 +26,9 @@
 
 @dataclass(frozen=True)
 class DependencyOverride:
-    analysis_id: str
+    from onecodex.models.analysis import Analyses
```
jairideout, line 29: nit: move import to top of file if possible

onecodex/cli.py
```diff
@@ -679,28 +679,29 @@ def jobs_group():
...
     parsed_dependencies = []
     for dep in dependency_overrides:
         analysis_id, _, download_path = dep.partition("=")
         if not analysis_id:
             raise click.BadParameter(
-                "Expected <analysis_id> or <analysis_id>=<download_path>, got {!r}.".format(dep),
+                f"Expected <analysis_id> or <analysis_id>=<download_path>, got {dep!r}.",
                 param_hint="-d/--dependency-override",
             )
+        dep_analysis = job = ctx.obj["API"].Analyses.get(analysis_id)
```
jairideout: why are we assigning to `dep_analysis` *and* `job`?

onecodex/cli.py
```diff
@@ -712,5 +713,5 @@ def jobs_run(ctx, job_id, sample_id, args, dependency_overrides, populate_defaul
         dependency_overrides=parsed_dependencies or None,
         populate_default_arguments=populate_default_arguments,
     )
-    click.echo(f"Job run successfully. New workflow ID: {run.id}")
-    click.echo(f"Get its status using `onecodex workflows {run.id}`")
+    click.echo(f"Job run created successfully. New analysis ID: {run.id}")
+    click.echo(f"Get its status using `onecodex analyses {run.id}`", err=True)
```
jairideout, line 717: Why `stderr`?

jairideout (approved):

onecodex/models/misc.py
pkonieczny: `onecodex.models.analysis` module already imports this file top-level, it's better not to create those circular references

onecodex/cli.py
pkonieczny: Throughout this file we sent comments to stderr - the way I see it, the line with ID is the output and the other one is just a comment 🤷

onecodex/cli.py
pkonieczny: Oops, that's a bug! Fixed

onecodex/models/analysis.py
pkonieczny: I felt it's a bit out of scope, but I guess it did pile on to the existing tech debt.
I refactored how `update` works in the client lib which allowed for skipping this part. `Analyses.refresh` now works without adding "update" to `_allowed_methods`

pkonieczny: See the comment around "update" method below

jairideout (approved):
Nice!

alenajk (approved):

PR #648, opened 2026-06-02
Add `await_completion` to `ocx.Analyses`

audy:
## Status
- [x] **Ready**

## Description

- Add `analysis.await_completion()` to poll for an analysis
- Add `onecodex await` to CLI
- Add `--await` to `onecodex jobs run`
- Move `ocx.Analyses.refresh` to `ocx.AnalysesBase.refresh` so that all analysis subclasses have it
- Add "Running Workflows" section to `README.md`
- Remove `ONE_CODEX_API_KEY` and `ONE_CODEX_BEARER_TOKEN` from environment when running tests

I snuck these two fixes in as well:

- `onecodex jobs run` now prints API error messages
- `onecodex.Jobs.run` now raises a `OneCodexException` if the API response is not okay. The exception's message includes the API response (useful for surfacing argument validation errors)

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Tests

### DIFF
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -341,6 +341,54 @@ def analyses(ctx, analyses):
     cli_resource_fetcher(ctx, "analyses", analyses)
 
 
+@onecodex.command("await")
+@click.argument("analysis_id", nargs=1, required=True)
+@click.option(
+    "--timeout",
+    type=float,
+    default=None,
+    help="Maximum number of seconds to wait. Waits indefinitely if not set.",
+)
+@click.option(
+    "--initial-interval",
+    type=float,
+    default=3.0,
+    show_default=True,
+    help="Seconds between the first polls.",
+)
+@click.option(
+    "--max-interval",
+    type=float,
+    default=120.0,
+    show_default=True,
+    help="Upper bound on the polling interval after backoff.",
+)
+@click.pass_context
+@pretty_errors
+@telemetry
+@login_required
+def analyses_await(ctx, analysis_id, timeout, initial_interval, max_interval):
+    """Poll an analysis until it reaches a terminal state."""
+    analysis = ctx.obj["API"].Analyses.get(analysis_id)
+    if not analysis:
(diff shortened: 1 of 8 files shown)


onecodex/cli.py
```diff
@@ -341,6 +341,59 @@ def analyses(ctx, analyses):
...
+@login_required
+def analyses_await(
+    ctx, analysis_id, timeout_seconds, initial_interval_seconds, max_interval_seconds
+):
+    """Poll an analysis until it reaches a terminal state."""
+    analysis = ctx.obj["API"].Analyses.get(analysis_id)
+    if not analysis:
+        raise click.ClickException(f"Could not find analysis {analysis_id} (404 status code)")
+
+    try:
```
pkonieczny, line 407: Would be cool to show some sort of spinner here

pkonieczny (approved):

onecodex/cli.py
```diff
@@ -341,6 +341,59 @@ def analyses(ctx, analyses):
     cli_resource_fetcher(ctx, "analyses", analyses)
 
 
+@onecodex.command("await")
```
jairideout: What do you think about grouping `await` under `analyses`? `onecodex analyses await`

jairideout (approved):

onecodex/cli.py
audy: @jairideout I agree that's the most natural place for it but `onecodex analyses` is already a command and I can't preserve backwards compatibility with that unless I do something kinda hacky with click.

onecodex/cli.py
audy: @pkonieczny it would be cool but I didn't want to roll my own spinner because it would produce a lot of output in a pipe. You can implement it in a way that detects TTY but I think that warrants its own PR (there's probably other places where we could add a spinner as well).

onecodex/cli.py
audy: @jairideout I decided that nesting now is probably better. We can replace it with a less hacky solution in a future major version release.

PR #650, opened 2026-06-05
Add persistent cache for analysis results

audy:
## Status

- [x] **Ready**

## Description

Adds a persistent, disk-based cache for fetching API results. The cache key includes `updated_at` so updated analysis runs will be re-fetched.

The cache is currently opt-in and off-by-default. The cache directory can be configured and defaults to the system's tempdir.

```python
import onecodex
ocx = onecodex.Api(cache_results=True) # uses system tempdir by default

# results will be cached on disk. If results from a previous session are found; results will be read from disk rather than re-downloading from remote.
ocx.Analyses.get("...").results()
```

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Tests
- [x] Documentation


petraszd (approved):

onecodex/api.py
```diff
@@ -182,6 +227,17 @@ def __init__(
             headers["X-OneCodex-Api-Experimental"] = "1"
 
         self._client = HTTPClient(auth=auth, headers=headers)
+
+        env_cache = os.environ.get("ONECODEX_DISK_CACHE")
```
jairideout: This should be `ONE_CODEX_DISK_CACHE` to match the other env vars (`ONE_CODEX_API_KEY`, `ONE_CODEX_BEARER_TOKEN`, `ONE_CODEX_AUTO_TELEMETRY`)

onecodex/cache.py
```diff
@@ -0,0 +1,156 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def _default_cache_dir() -> Path:
+    # Default to a per-user dir under the system tempdir so the cache doesn't grow
+    # unboundedly on long-lived machines: tempdirs are reclaimed across reboots
+    # (launchd on macOS, systemd-tmpfiles on Linux). Callers that want persistence
+    # across reboots should pass an explicit path instead.
+    return Path(tempfile.gettempdir()) / f"onecodex-results-cache-{os.getuid()}"
```
jairideout: `os.getuid` only works on Unix and doesn't work in WASM. Is there something we could use that's more portable, e.g. the logged in user's ID? That might work better, e.g. if a user caches some results logged in as `user A`, logs out, and then logs in as `user B`

onecodex/cache.py
```diff
@@ -0,0 +1,156 @@
...
+            log.warning("onecodex disk cache: dropping corrupt entry %s (%s)", entry, e)
+            try:
+                entry.unlink()
+            except FileNotFoundError:
+                pass
+            return None
+
+    def set(self, key: str, value: Any) -> None:
+        entry = self._entry_path(key)
+        tmp = entry.with_suffix(f".tmp.{os.getpid()}.{secrets.token_hex(4)}")
```
jairideout, line 123: nit: can we just overwrite the existing file for simplicity, rather than creating a temp file and replacing the original?

onecodex/cache.py
audy: I switched to `getuser`. Using the OCX user is a good idea but more complicated to implement. Currently, `.results()` does not change depending on who is accessing it.

onecodex/cache.py
audy: If the process is killed then we'll end up with a partially complete file. `.get` will handle it automatically but I think it's probably worth avoiding in the first place.

audy:
There was a simpler way

PR #655, opened 2026-06-10
Add `--sample` to `onecodex download samples`

audy:
## Status

- [x] **Ready**

## Description

Adds `--sample-id` flag to `onecodex download samples`. The flag can be repeated to fetch more than one sample by ID. Cannot be used with `--project` or `--tags` (it does not make sense to filter by project and/or tags when specifying exact sample ID(s)).

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Documentation

### DIFF
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -557,7 +557,7 @@ def download_group():
 )
 @click.option(
     "-s",
-    "--sample-id",
+    "--sample",
     "sample_ids",
     multiple=True,
     help="Download a specific sample by its ID. Can be passed multiple times. Mutually "
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -556,6 +556,14 @@ def download_group():
     "-t", "--tags", multiple=True, help="Filter to samples that include *any* of these tag names."
 )
 @click.option(
+    "-s",
+    "--sample-id",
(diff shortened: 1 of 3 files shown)


onecodex/lib/download.py
```diff
@@ -50,11 +50,21 @@ def download_samples(
...
+    sample_ids=None,
     prompt=False,
     min_samples_for_prompt=50,
     progressbar=False,
 ):
-    if project_name_or_id:
+    if sample_ids:
+        samples = []
+        for sample_id in sample_ids:
+            log.info("Fetching sample '{}'...".format(sample_id))
```
jairideout: nit: use f-strings (here and below)

onecodex/cli.py
```diff
@@ -555,6 +555,14 @@ def download_group():
 @click.option(
     "-t", "--tags", multiple=True, help="Filter to samples that include *any* of these tag names."
 )
+@click.option(
+    "-s",
+    "--sample-id",
```
jairideout: `--sample` might be more consistent with `--project`/`--tags`. It'd also allow us to select samples by other attributes in the future (e.g. filename, name, etc.)

jairideout (approved):

PR #677, opened 2026-07-14
Fix to_functional_df function

pkonieczny:
## Status

- [X] **Ready**

## Description

This PR makes 2 major changes:
1. It changes the shape of `to_functional_df` to match `to_classification_df`: functional profiles (so, samples) are the rows whereas function ids (or function ids x tax ids for stratified results) are the columns. For stratified results, the function ids x tax ids is a MultiIndex on the columns.
2. It ensures that `taxa_stratified=True` is handled properly in that data frame: each contributing taxon gets its own column in the dataframe.

As part of this PR some benchmarks were run on sparse representation of the data frame - the verdict is: not worth it since the wide format introduces _a lot_ of columns so huge overhead in pandas. The sparse format actually uses more memory due to that even at values at ~25% full. We'd need ~1k samples for sparse to become useful.

## Related PRs

- [X] This PR is independent

## TODOs

- [ ] Update `CHANGELOG` ?


onecodex/models/base_sample_collection.py
```diff
@@ -52,6 +53,18 @@ class MetadataFetchResults:
...
 
+def _normalize_taxon_field(value: Any) -> str:
+    """Coerce a taxon id/name to a string for use in a DataFrame index.
+
+    Required to consistently handle missing values, ints and floats (like 386414.0).
+    """
+    if value is None or (isinstance(value, float) and math.isnan(value)):
+        return ""
+    if isinstance(value, float) and value.is_integer():
+        return str(int(value))
```
petraszd, line 64: Can it contain float discrepancies like `5.999999999999999`?

petraszd (approved):
Code changes looks good. Though I am not fully sold on breaking backwards compatibility.

onecodex/models/base_sample_collection.py
pkonieczny: This is specifically for tax id values - we don't do any arithmetic on them so we shouldn't get a value like that. It's for when sometimes pandas will auto-convert the int-like values to e.g. `431.0`

treynr (approved):

PR #680, opened 2026-07-23
Handle colons in Altair chart field names

jairideout:
## Status

- [x] **Ready**

## Description

Metadata field names containing a colon (e.g. "foo:bar") were passed to Altair as encoding shorthands and raised a ValueError.

`escape_chart_fields()` now sets field names explicitly with `field=`, which skips shorthand parsing.

Closes DEV-11880

## Related PRs

- [x] This PR is independent

### DIFF
--- a/onecodex/viz/_primitives.py
+++ b/onecodex/viz/_primitives.py
@@ -1,9 +1,15 @@
+from __future__ import annotations
+
 import os
 from itertools import chain
 from collections.abc import Iterable
 from math import ceil
+from typing import TYPE_CHECKING
 
 from onecodex.exceptions import OneCodexException
+
+if TYPE_CHECKING:
+    import altair as alt
 
 
 # We use custom IDs in some databases and we don't want to link them to NCBI. They are always after 2e9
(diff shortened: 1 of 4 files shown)


alenajk (approved):

onecodex/viz/_primitives.py
```diff
@@ -164,6 +164,28 @@ def get_unique_column(preferred_name, existing_columns):
...
+    field, sep, maybe_type = shorthand.rpartition(":")
+    if sep and maybe_type in _ALTAIR_TYPECODES:
+        suffix = sep + maybe_type
+    else:
+        field, suffix = shorthand, ""
+    field = field.replace(".", r"\.").replace("[", r"\[").replace("]", r"\]").replace(":", r"\:")
+    return field + suffix
+
+
 def escape_chart_fields(chart):
```
petraszd: For now it is fine. Or it is better than previous implementation. But I am not sure if it is leaky or not.

----

(_Feel free to ignore the rest. It is just a rant_)

This doc:
https://altair-viz.github.io/user_guide/encodings/index.html#escaping-special-characters-in-column-names

It is seems that escaping `:` is leaky. So I wonder if we should just go through all the code and make sure all fields are defined using explicitly (`ClassName(field=)`) instead of using shorthands everywhere.

petraszd (approved):

onecodex/viz/_primitives.py
pkonieczny: The docs suggest that the escape character is actually displayed in the resulting chart - if that's true, I think that replacing with underscore (or similar) may actually look better

pkonieczny (approved):

onecodex/viz/_primitives.py
jairideout: Updated to use explicit `field=`; we don't have to escape or substitute characters anymore

alenajk (approved):

PR #684, opened 2026-08-20
Passing non-exising `group_by` to `.plot_bargraph` should not throw error

petraszd:
## Status

- [x] **Ready**

## Description

`haxis` param also accepts metadata fields but it does not throw an error if field does not exist. `group_by` should act the same.

Other option would be to convert `OneCodexException` to `PlottingException`, but it still would be inconsistent with `haxis`.

## Related PRs

- [x] This PR is independent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `onecodex jobs publish` CLI command.
 - Added `description` argument to `Jobs.run()`.
 - Added `--description` option to `onecodex jobs run` CLI command.
-- Fixed `plot_bargraph`: passing non-existing `group_by` should not throw an error.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `onecodex jobs publish` CLI command.
 - Added `description` argument to `Jobs.run()`.
 - Added `--description` option to `onecodex jobs run` CLI command.
+- Fixed `plot_bargraph`: passing non-existing `group_by` should not throw an error.
```
jairideout: nit: maybe this fits better under the "Fixed" heading below?

onecodex/viz/_bargraph.py
```diff
@@ -139,6 +136,10 @@ def plot_bargraph(
         if group_by:
             # calculate per-group mean of each taxon
 
+            # In case `group_by` does not exist
+            if group_by not in self.metadata:
+                self.metadata[group_by] = np.nan
```
jairideout: I don't think we should mutate `self.metadata` in-place, because it can be used outside of this method and might surprise the user to see a new column added

petraszd: Great catch. Thank you!

Fixed.

pkonieczny (approved):

jairideout (approved):

PR #686, opened 2026-08-21
Optimise the subset reads script

petraszd:
## Status

- [x] **Ready**

## Description

- Speeding up subset reads script
- Adding gzip'ing option. It is faster to gzip while writing then write and call gzip afterwards.

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Update `CHANGELOG`

### DIFF
--- a/onecodex/scripts/subset_reads.py
+++ b/onecodex/scripts/subset_reads.py
@@ -52,14 +52,17 @@ def validating_parser(file_path, **io_kwargs):
         yield buf.read()
 
 
-def get_filtered_filename(file_path):
+def get_filtered_filename(file_path, gzip_output):
     filename = os.path.basename(file_path)
     filename, ext = os.path.splitext(filename)
 
     if ext in {".gz", ".gzip", ".bz", ".bzip", ".bz2", ".bzip2"}:
         filename, ext = os.path.splitext(filename)
 
-    return "{}.filtered{}".format(filename, ext), ext
+    out_filename = "{}.filtered{}".format(filename, ext)
+    if gzip_output and not out_filename.endswith(".gz"):
+        out_filename += ".gz"
+    return out_filename, ext
 
 
 def make_taxonomy_dict(classification, parent=False):
@@ -294,15 +297,11 @@ def cli(
         tsv_row_count = int(tsv_row_count / 2.0)
 
     # determine the name of the output file(s)
-    filtered_filename, ext = get_filtered_filename(fastx)
+    filtered_filename, ext = get_filtered_filename(fastx, gzip_output)
     filtered_filename = os.path.join(out, filtered_filename)
-    if gzip_output and not filtered_filename.endswith(".gz"):
-        filtered_filename += ".gz"
     if reverse:
-        rev_filtered_filename = get_filtered_filename(reverse)[0]
+        rev_filtered_filename = get_filtered_filename(reverse, gzip_output)[0]
         rev_filtered_filename = os.path.join(out, rev_filtered_filename)
-        if gzip_output and not rev_filtered_filename.endswith(".gz"):
-            rev_filtered_filename += ".gz"
 
     if ext in {".fa", ".fna", ".fasta"}:
         io_kwargs = {"format": "fasta"}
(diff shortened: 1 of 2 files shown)


onecodex/scripts/subset_reads.py
```diff
@@ -179,6 +200,17 @@ def too_many_fastx_records():
     help="By default, reads with low confidence taxonomic assignments are ignored. "
     "Choose this option to include them.",
 )
+@click.option(
+    "--gzip-output",
```
pkonieczny, line 207: Could we just look at the extension and gzip if user provides a `X.gz` output file?

petraszd: Hmmm... Explicit is better than implicit.

pkonieczny: We already do that in fastx validator 🤷

onecodex/scripts/subset_reads.py
```diff
@@ -280,8 +318,16 @@ def cli(
     # see mainline/#3513. we must set idx=0 here for cases where the fastx file is empty
     idx = 0
 
-    with click.progressbar(length=tsv_row_count) as bar, gzip.open(readlevel_path, "rt") as tsv:
-        reader = csv.DictReader(tsv, delimiter="\t")
+    # with click.progressbar(length=tsv_row_count) as bar, gzip.open(readlevel_path, "rt") as tsv:
```
pkonieczny: Why keep the commented out code?

pkonieczny (approved):

onecodex/scripts/subset_reads.py
petraszd: I've forgot to delete it.

alenajk (approved):

onecodex/scripts/subset_reads.py
```diff
@@ -280,8 +318,15 @@ def cli(
     # see mainline/#3513. we must set idx=0 here for cases where the fastx file is empty
     idx = 0
 
-    with click.progressbar(length=tsv_row_count) as bar, gzip.open(readlevel_path, "rt") as tsv:
-        reader = csv.DictReader(tsv, delimiter="\t")
+    with click.progressbar(length=tsv_row_count) as bar, gzip.open(readlevel_path, "rb") as tsv:
+        # Using `csv.DictReader(tsv, delimiter="\t")` would be more readable but doing raw parsing is
```
jairideout, line 325: How do these approaches compare?

- `csv.reader`
- `pandas.read_csv(..., sep="\t")`

jairideout (approved):

onecodex/scripts/subset_reads.py
audy: pandas is going to read the entire TSV into memory so I wouldn't do that. Also `pandas` isn't in the base dependencies.

onecodex/scripts/subset_reads.py
audy: There is no way for the user to provide the exact output filename (`--out` is a directory, not a path).

The help string for `--out` is ambiguous, @petraszd do you mind updating it in this PR? "Directory where output file(s) will be saved"

Also, maybe `type` should actually be? `click.Path(exists=True, file_okay=False, dir_okay=True)`

onecodex/scripts/subset_reads.py
```diff
@@ -258,9 +292,13 @@ def cli(
     # determine the name of the output file(s)
     filtered_filename, ext = get_filtered_filename(fastx)
     filtered_filename = os.path.join(out, filtered_filename)
+    if gzip_output and not filtered_filename.endswith(".gz"):
+        filtered_filename += ".gz"
     if reverse:
         rev_filtered_filename = get_filtered_filename(reverse)[0]
         rev_filtered_filename = os.path.join(out, rev_filtered_filename)
+        if gzip_output and not rev_filtered_filename.endswith(".gz"):
```
audy: nit: I would just add an extension argument to `get_filtered_filename`

audy (approved):
The bottleneck ~still seems to be~ is now fastx parsing. How much faster would this be if we used needletail instead? It could be added as an optional dependency (since it's compiled, not all platforms have it)

onecodex/scripts/subset_reads.py
petraszd: Done.

PR #687, opened 2026-08-24
Support compact functional results

treynr:
## Status

- [x] **Ready**

## Description

This PR adds support for fetching functional results directly from the URL provided by `results_uri`. When this static file is unavailable, results are fetched via the API as a fallback. Because this static results file is structured differently, the results are transformed to be compatible with the current API. 

This PR includes a number of changes:
- Make S3 and results loading functions generic to use with both `Classifications` and `FunctionalProfiles`
- Add results formatting (rehydration) functions to maintain API compatibility
- Add support for static file fetching when using `.results()`, `.table()`, or `.to_functional_df()`
- Add mocked data and test cases for using `results_uri` file

## Related PRs

- [x] This PR is independent


onecodex/models/analysis.py
```diff
@@ -555,12 +799,52 @@ def where(  # type: ignore[override]
 class FunctionalProfiles(_AnalysesBase, FunctionalRunSchema):
     _resource_path = "/api/v1/functional_profiles"
 
+    _FUNCTIONAL_RESULTS_VERSION = 1
+
+    @lru_cache
```
pkonieczny: Caching the full functional results (even condensed) in memory is quite dangerous, I'm not sure we want to do it

onecodex/models/analysis.py
```diff
@@ -53,6 +51,267 @@ def _decompress(data: bytes) -> bytes:
...
+            "pathways",
+            pathway_id,
+            pathway_name,
+            "abundance",
+            community_abundance,
+            destination=pathway_table,
+        )
+
+        for taxon_id, abundance, coverage, cpm in contributions:
+            # todo: to keep strict compatibility with current results we set the
```
pkonieczny: I don't think we need strict compatibility here, especially that we're breaking compatibility anyway

onecodex/models/analysis.py
```diff
@@ -53,6 +51,267 @@ def _decompress(data: bytes) -> bytes:
...
+    hydrated = _rehydrate_functional_results(filtered_condensed_results)
+
+    table = []
+    for row in hydrated["table"]:
+        if (
+            row["group_name"] != annotation.value
+            or row["metric"] != effective_metric.value
+            or row["taxa_stratified"] != taxa_stratified
+        ):
+            continue
```
pkonieczny: How much more complex would it be to skip other annotations/metrics before or during hydration instead?

tests/test_functional_profiles.py
```diff
@@ -264,3 +266,240 @@ def test_filter_functional_runs_to_newest_job(ocx, raw_api_data, custom_mock_req
...
+
+    assert profile.results_uri is not None
+
+    condensed = profile._condensed_results()
+    assert condensed is not None
+
+    actual = profile.results()
+
+    # the results directly from the API endpoint
+    response = profile._client.get(f"{profile._api._base_url}{profile.field_uri}/results")
```
pkonieczny: The results from the API endpoint will be condensed as well, we should maybe keep around a test data file with original API results to compare to

tests/test_functional_profiles.py
```diff
@@ -264,3 +266,240 @@ def test_filter_functional_runs_to_newest_job(ocx, raw_api_data, custom_mock_req
...
+        fill_missing=False,
+    )
+
+    # Metadata may be fetched, but functional results must not come from
+    # the filtered-results endpoint.
+    new_requests = api_data.calls[request_count:]
+    assert all("/filtered_results" not in call.request.url for call in new_requests)
+
+    response = profile._client.get(
+        f"{profile._api._base_url}{profile.field_uri}/filtered_results",
```
pkonieczny: We want to get rid of that endpoint altogether, we should use stored test data files instead

pkonieczny (commented):
Overall LGTM, but in tests we shouldn't compare against endpoint API results I think

onecodex/models/analysis.py
```diff
@@ -53,6 +51,267 @@ def _decompress(data: bytes) -> bytes:
     return data
 
 
+@lru_cache(maxsize=1)
+def _get_s3_session():
+    """Return a requsets session that can be used for retrieving results from S3 using presigned URLs."""
```
audy: typo: requsets

onecodex/models/analysis.py
audy: +1 for filtering upstream (or skipping rehydration altogether)

audy (commented):
Removing the rehydration step and updating `.to_functional_df` to take the condensed format instead is simpler (which is a big win IMO because keeping track of all of the different formats and transformations is already difficult), and is probably considerably more efficient.

`to_functional_df` would still be reverse-compatible (so downstream stuff like plotting wouldn't break).

onecodex/models/analysis.py
audy: Let's skip caching for now* and add disk caching in a separate PR

treynr:
> Removing the rehydration step and updating `.to_functional_df` to take the condensed format instead is simpler (which is a big win IMO because keeping track of all of the different formats and transformations is already difficult), and is probably considerably more efficient.
> 
> `to_functional_df` would still be reverse-compatible (so downstream stuff like plotting wouldn't break).

Yeah, it would be simpler. One concern is that removing the rehydration step would change the schema/shapes of `.results()` and `.table()`. Are we okay with that?

tests/conftest.py
```diff
@@ -64,25 +67,80 @@ def update_metadata_callback(req):
...
-            "value": elem["value"],
-            # Stratified results have additional "taxon_id" and "taxon_name" fields
+            "id": row["id"],
+            "name": row["name"],
+            "value": row["value"],
             **(
-                {"taxon_id": elem["taxon_id"], "taxon_name": elem["taxon_name"]}
+                {
+                    "taxon_id": row["taxon_id"],
+                    "taxon_name": row["taxon_name"],
```
pkonieczny, line 106: Is taxon_name included in the results?

tests/test_functional_profiles.py
```diff
@@ -264,3 +266,231 @@ def test_filter_functional_runs_to_newest_job(ocx, raw_api_data, custom_mock_req
...
+    comparison_keys = ("id", "name", "value")
+
+    if taxa_stratified:
+        expected_keys |= {"taxon_id", "taxon_name"}
+        comparison_keys += ("taxon_id",)
+
+    assert all(set(row) == expected_keys for row in actual["table"])
+    assert all(set(row) == expected_keys for row in expected["table"])
+
+    def normalize(rows):
```
pkonieczny: It seems this function could be extracted as it has a very similar definition above

onecodex/models/analysis.py
```diff
@@ -53,6 +51,284 @@ def _decompress(data: bytes) -> bytes:
...
+
+        data = resp.content
+    else:
+        with open(uri, "rb") as f:
+            data = f.read()
+
+    return orjson.loads(_decompress(data))
+
+
+def _rehydrate_functional_results(
```
pkonieczny: Nit: that's a lot of code, maybe we could put it in a separate `functional.py` file?

onecodex/models/analysis.py
```diff
@@ -53,6 +51,284 @@ def _decompress(data: bytes) -> bytes:
...
+    else:
+        selected_results = {
+            annotation.value: [
+                feature
+                for feature in condensed_results["results"].get(annotation.value, [])
+                if feature[0] not in _SKIP_FUNCTIONAL_IDS
+            ]
+        }
+
+    # shouldn't mutate the original otherwise we change what @lru_cache._condensed_results()
```
pkonieczny: Nit: there's no longer a `lru_cache` on `_condensed_results()`

pkonieczny (approved):
I like Austin's idea of creating `to_functional_df` from condensed results and maybe exposing the "old" results format via some other function - but this also works!

LGTM!

tests/conftest.py
treynr: yeah, for `/filtered_results` it is.

PR #692, opened 2026-09-04
Merge functional table functions and simplify filtered_table/rehydration

treynr:
## Status

- [x] **Ready**

## Description

This PR merges the `table()` and `filtered_table()` functions and alters `filtered_table()`/`to_functional_df()` to use condensed results instead of pulling from the `/filtered_results` endpoint. `table()` has been updated to include a new `metric` argument, allowing it to generate filtered results. `filtered_table()` is kept for backwards compatibility but is now a wrapper around `table()`. This PR also removes usage of the `/filtered_results` endpoint.

Summary of changes:
- generate `to_functional_df()` using condensed results
- gets rid of the filtered rehydration function
- add metric arg to `table()` and make `filtered_table()` a wrapper
- remove `/filtered_results` endpoint calls
- update the custom plots functional mock
- remove `/filtered_results` tests and add tests for the new behavior

## Related PRs

- [x] This PR is independent


onecodex/models/analysis.py
```diff
@@ -646,6 +637,7 @@ def table(
         self,
         annotation: Optional[FunctionalAnnotations] = None,
         taxa_stratified: bool = True,
+        metric: FunctionalAnnotationsMetric | Literal["all"] = "all",
```
audy, line 640: I think this also needs `| str`

onecodex/models/analysis.py
```diff
@@ -664,41 +659,51 @@ def table(
...
+        empty_df = pd.DataFrame(
+            {
+                "group_name": pd.Series(dtype="str"),
+                "id": pd.Series(dtype="str"),
+                "name": pd.Series(dtype="str"),
+                "metric": pd.Series(dtype="str"),
+                "value": pd.Series(dtype="float"),
+                "taxa_stratified": pd.Series(dtype="bool"),
+                "taxon_id": pd.Series(dtype="str"),
+                "taxon_name": pd.Series(dtype="str"),
```
audy, line 673: should we drop the greengenes-style taxonomies now?

onecodex/models/functional.py
```diff
@@ -37,10 +44,12 @@ def add_row(
         destination: Optional[list] = None,
     ) -> None:
         """Add a row formatted for results['table']."""
-        # no need to rehydrate rows we don't need
+        # no need to rehydrate rows we don't need, se skip certain functional ids to maintain
```
audy, line 47: typo? "se"

onecodex/models/functional.py
```diff
@@ -176,81 +188,157 @@ def add_species_rows(
...
-    annotation: FunctionalAnnotations,
-    metric: FunctionalAnnotationsMetric,
+    annotation: FunctionalAnnotations | str,
+    metric: FunctionalAnnotationsMetric | str,
     taxa_stratified: bool,
 ) -> dict:
-    """Rehydrate condensed functional results into the public filtered API results format."""
+    """Select one metric directly from condensed functional results.
+
+    This does not construct legacy result-row dictionaries or an intermediate
```
audy, line 237: This comment no longer seems necessary (editorializing)

onecodex/viz/_custom_plots/collection.py
```diff
@@ -116,20 +116,40 @@ def _filtered_results(
...
         if taxa_stratified:
             raise OneCodexException("Taxa stratified results are not currently supported")
 
-        return {
-            "table": self._results.get(f"{annotation}-{metric}", []),
-            "n_reads": self._results["n_reads"],
-            "n_mapped": self._results["n_mapped"],
-        }
+        return _select_condensed_functional_results(
+            self._results,
```
audy, line 126: is `self._results` still coming from `/api/frontend/custom-plots/functional-results`? If so, will that format work with `_select_condensed_functional_results`?

audy (changes requested):
Looks good overall. See comment about custom plots endpoint. Maybe I missed something. But I think we should do some e2e QA before merging either way.

onecodex/models/analysis.py
treynr: they're dropped. this `taxon_name` is now a proper binomial name that's pulled from the taxonomy associated with the results.

audy: It's nicer to have the taxon name but since classification results don't, I wonder if we should exclude them here as well (to be consistent).

onecodex/models/analysis.py
```diff
@@ -664,41 +659,50 @@ def table(
...
+            result_json,
+            annotation_filter=annotation,
+            metric_filter=metric if metric != "all" else None,
+            taxa_stratified_filter=taxa_stratified,
+        )
+
+        if not result_json["table"]:
+            return empty_df
+
+        return pd.DataFrame(result_json["table"], columns=empty_df.columns)
```
audy: I would maybe do this so that the column types are always consistent:

```python

# drop if the if not results_json['table'] early return

METRICS_SCHEMA = {
    "group_name": "string",
    "id": "string",
    "name": "string",
    "metric": "string",
    "value": "float64",
    "taxa_stratified": "boolean",
    "taxon_id": "string",
}

return pd.DataFrame(results_json['table'] or [], columns=list(METRICS_SCHEMA)).astype(METRICS_SCHEMA)

```

audy: Or this, so that pandas doesn't coerce twice:

```python
    return pd.DataFrame(
        {
            name: pd.Series([r.get(name) for r in results_json['table'] or []], dtype=dtype)
            for name, dtype in METRICS_SCHEMA.items()
        }
    )
```

onecodex/models/analysis.py
```diff
@@ -596,33 +596,24 @@ def _results(self) -> dict:
...
+        condensed_results = self._condensed_results()
 
-            return _rehydrate_filtered_functional_results(
-                condensed_results,
-                annotation=annotation,
-                metric=metric,
-                taxa_stratified=taxa_stratified,
-            )
+        if condensed_results is None:
+            # shouldn't get here
```
audy: If the pre-signed URL expires you will get here. Falling back to `/results` (like classification runs) won't work yet.

Maybe we can re-fetch the functional run to update the URI?

onecodex/models/functional.py
```diff
@@ -176,81 +188,154 @@ def add_species_rows(
...
+        [id, name, community_cpm, total_rpk, contributions]
+
+    where each contribution is:
+
+        [taxon_id, cpm, rpk]
 
-    _SKIP_FUNCTIONAL_IDS = {"UNMAPPED", "UNGROUPED", "UNINTEGRATED"}
+    Pathway rows have the form:
+
+        [
```
audy, line 247: It might be worth using a `NamedTuple` for these. Then you don't need to rely on the docstring + list indices to remember which position is which field.

This can be in a later PR though because I think it is all internal.

audy (approved):
Some minor nits + the CHANGELOG needs an update

PR #693, opened 2026-09-07
Add filtering samples by tax ids

pkonieczny:
## Status

- [X] **Ready**

## Description

This PR adds filtering samples by tax ids. **Important**: wait until the corresponding mainline PR is merged and deployed.
Also in this PR: fixing metadata filtering on samples where _both_ sample and metadata filters were present. Adding a bunch of tests to ensure all cases are covered and work as expected.

## Related PRs

- [X] This PR is independent

## TODOs

- [x] Merge sample taxa filter API v1 PR first

### DIFF
--- a/onecodex/models/sample.py
+++ b/onecodex/models/sample.py
@@ -318,7 +318,7 @@ class Samples(OneCodexBase, _SampleSchema, ResourceDownloadMixin):
                     raise OneCodexException("Unknown tag specified: {}".format(t))
             new_tags.append(new_tag)
 
-        # Merge inline field kwargs with leftover **keyword_filters.
+        # merge inline field kwargs with leftover **keyword_filters
         merged_filters = _drop_unset(
             created_at=created_at,
             updated_at=updated_at,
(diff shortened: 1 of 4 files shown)


jairideout (approved):

petraszd (approved):

onecodex/models/sample.py
```diff
@@ -204,7 +205,8 @@ def where(  # type: ignore[override]
         public: bool = False,
         organization: bool = False,
         filter: Any = None,
-        tags: list | None = None,
+        tags: list[str] | None = None,
```
audy, line 208: I think this can be a `list[str, Tags, dict]`?

This works:

```python
ocx.Samples.where(tags=[ocx.Tags.get('2bbf3f5a6992436f')], limit=10)
```

audy: Actually maybe it's `str | list[str, Tags, dict]`:

```
ocx.Samples.where(tags='high host', limit=10)
```

onecodex/models/sample.py
```diff
@@ -238,6 +240,10 @@ def where(  # type: ignore[override]
 
             ocx.Samples.where(tags=["trimmed", "human-depleted"])
 
+        Filter by tax ids (returns samples containing *all* the listed taxa)::
```
audy, line 243: Can you add that "contains" here means one or more cumulative reads (corresponding to `Metric.ReadcountWChildren`)?

onecodex/models/sample.py
```diff
@@ -334,16 +345,24 @@ def where(  # type: ignore[override]
             for k, v in merged_filters.items()
             if k in Metadata.model_fields and k not in Samples.model_fields
         }
-        if not public and not organization:
-            if md_search_keywords:
-                metadata_samples = [md.sample for md in Metadata.where(**md_search_keywords)]
+        # drop Metadata filters from Samples filters
```
audy, line 348: I think this fixes DEV-12014

audy (approved):
Awesome!

onecodex/models/sample.py
pkonieczny: Good point, updated the type

onecodex/models/sample.py
pkonieczny: It does - I missed that ticket! I've added a regression test to make sure

onecodex/models/sample.py
pkonieczny: Added

onecodex/models/sample.py
pkonieczny: I won't reference the ticket though as it seems it also has to do with enum partial matching

PR #695, opened 2026-09-08
CLI commands to list and download an analysis' output files

jairideout:
## Status

- [x] **Ready**

## Description

- `onecodex analyses files ID` lists output files (tabular or JSON format)
- `onecodex analyses download ID` saves output files under `-o/--out`. A subset of output files can be downloaded via `-f/--file`
- `onecodex analyses <subcommand> --help` (e.g. `onecodex analyses logs --help`) no longer requires being logged in

Closes DEV-11796

## Related PRs

- [x] This PR is independent

### DIFF
--- a/onecodex/cli.py
+++ b/onecodex/cli.py
@@ -25,6 +25,8 @@ from onecodex.input_helpers import (
 )
 from onecodex.lib.upload import DEFAULT_THREADS
 from onecodex.metadata_upload import validate_appendables
+from onecodex.models import Analyses
+from onecodex.models.schemas.misc import FileDetailSchema
 from onecodex.scripts import export_functional_metric, interleave, subset_reads
 from onecodex.utils import (
     OPTION_HELP,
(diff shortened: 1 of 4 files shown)


onecodex/cli.py
```diff
@@ -464,17 +479,133 @@ def analyses_await(
...
+    if as_json:
+        pprint([x.model_dump() for x in files], ctx.obj["NOPPRINT"])
+        return
+
+    if not files:
+        click.echo(f"Analysis {analysis_id} has no output files.")
+        return
+
+    rows = [("Filepath", "Size (bytes)")] + [(x.filepath, str(x.size)) for x in files]
+    width = max(len(filepath) for filepath, _ in rows)
```
petraszd: Nit: what if filepath is too large?

----

We probably should just ignore it. Similarly how psql just prints out whatever you are selecting.

petraszd (approved):

onecodex/cli.py
```diff
@@ -464,17 +479,133 @@ def analyses_await(
...
+@click.option(
+    "--json", "as_json", is_flag=True, default=False, help="Output JSON instead of prettified table"
+)
+@click.pass_context
+@pretty_errors
+@telemetry
+@login_required
+def analyses_files(ctx: click.Context, analysis_id: str, as_json: bool) -> None:
+    """List the output files of an analysis."""
+    files = _get_analysis(ctx, analysis_id).get_files()
```
audy, line 497: why not use `cli_resource_fetcher`?

audy (approved):
Since we auto-tar the output when the number of files is huge, it'd be nice to have a flag to automatically un-tar. But that can be added later.

onecodex/cli.py
jairideout: `cli_resource_fetcher` resolves a top-level API model by name and calls `.all()` or `.get(id)` on it. Output files aren't a registered model (there's no `FileDetails` in `onecodex/models/__init__.py`)

PR #698, opened 2026-09-16
Expand `SampleCollection.filter` documentation

audy:
## Status

- [x] **Ready**

## Description

- Expand documentation for `SampleCollection.filter` with examples
- Add type annotations
- Minor correction: the return value of the callable passed to `.filter` does _not_ need to return a bool. `.filter` keeps all records where `filter_func(obj)` is _truthy_

## Related PRs

- [x] This PR is independent

## TODOs

- [x] Documentation


petraszd (approved):

onecodex/models/base_sample_collection.py
```diff
@@ -284,19 +284,53 @@ def filter(self, filter_func):
...
         Examples
         --------
-        Generate a new collection of Samples that have a specific filename extension:
+        Keep only the samples whose filename ends in ``.fastq.gz``::
 
-        >>> new_collection = samples.filter(lambda s: s.filename.endswith('.fastq.gz'))
+            samples = ocx.Samples.all(limit=10)
+            filtered = samples.filter(lambda s: s.filename.endswith('.fastq.gz'))
+
+        Note: For some fields including ``filename``, it's quicker and simpler to filter using
```
pkonieczny: Nit: I'd word this more strongly: the `where()` route should be prioritised where available

onecodex/models/base_sample_collection.py
```diff
@@ -284,19 +284,53 @@ def filter(self, filter_func):
...
+        :doc:`/querying` for more examples::
+
+            filtered = ocx.Samples.where(filename={'$endswith': '.fastq.gz'})
+
+        Custom metadata cannot be filtered using ``where``, so ``filter`` is necessary::
+
+            samples = ocx.Samples.where(project=project)
+            filtered = samples.filter(lambda s: s.metadata.custom.get('subject') == '123')
+
+        Keep only the samples that contain *Phocaeicola dorei* (tax id 357276)::
```
pkonieczny: This could be misleading to AI, I think we should have it always use the `Sample.where(tax_ids=["123"])` route. 
This comment makes it ambiguous.

pkonieczny (approved):
It seems AI works much better with less ambiguity - we should have opinionated instructions on which route to take. I'd reword the instructions to always have a clear single path, but that's just my opinion 🤷

jairideout (approved):

onecodex/models/base_sample_collection.py
audy: Removed.

pkonieczny (approved):

PR #704, opened 2026-09-24
use NamedTuples for functional group and pathway rows

alenajk:
## Status

- [x] **Ready**

## Description

Use NamedTuples instead of hard to read index numbers for functional group and pathway row members

## Related PRs

- [x] This PR is independent

## TODOs

- [ ] Tests
- [ ] Documentation
- [ ] Update `CHANGELOG`


onecodex/models/functional.py
```diff
@@ -288,16 +317,17 @@ def _select_condensed_functional_results(
 
     features = condensed_results["results"].get(results_group, [])
 
-    for feature in features:
-        feature_id = str(feature[0])
-        encoded_name = feature[1]
+    for row in features:
+        feature = feature_type._make(row)
```
petraszd: Is it worth performance hit? I think there are a lot of `features` and converting each row to a named tuple object inside a hot loop is not cheap. Maybe it is and maybe `features` list is small enough.

`feature.coverage` is not that much readable than `feature[coverage_idx]` but it may cost more.

I think it is worth it to do a performance testing on this.

onecodex/models/functional.py
```diff
@@ -312,19 +342,19 @@ def _select_condensed_functional_results(
 
         if not taxa_stratified:
             feature_ids.append(feature_id)
-            values.append(feature[community_value_index])
+            values.append(getattr(feature, value_field))
```
petraszd: Actually new line is less readable than previous one.

onecodex/models/functional.py
```diff
@@ -312,19 +342,19 @@ def _select_condensed_functional_results(
...
             feature_name_map[feature_id] = feature_name
             continue
 
         assert taxon_ids is not None
 
-        # Contributions are always the final element of a condensed feature.
-        for contribution in feature[-1]:
-            taxon_id = _normalize_taxon_id(contribution[0])
+        for raw_contribution in feature.contributions:
+            contribution = contribution_type._make(raw_contribution)
```
petraszd: Same here. It cost some CPU time to call `_make` in each iteration. Maybe it is enough to define `0` as `taxon_id_idx` and have `contribution[taxon_id_idx]`?

onecodex/models/functional.py
alenajk: @petraszd 
I had claude do some benchmarking with ~220k synthetic contributions and there was no significant difference in performance over a few runs of it

